PR #17, opened 2016-12-06
Enable FPU in MSTATUS CSR during init

scottj97:
This way, chips with FPU can run FPU instructions without taking
exceptions.

I've confirmed that dhrystone still runs successfully even on chips
with no FPU.

### DIFF
--- a/bsp/env/freedom-e300-hifive1/init.c
+++ b/bsp/env/freedom-e300-hifive1/init.c
@@ -182,6 +182,7 @@ void _init()
   printf("core freq at %d Hz\n", get_cpu_freq());
 
   write_csr(mtvec, &handle_trap);
+  write_csr(mstatus, MSTATUS_FS); // enable FPU
 
   _exit(main(0, NULL));
 }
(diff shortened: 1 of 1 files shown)


mwachs5:
Do we need to do something in the Makefile to take advantage of this?

scottj97:
No changes needed to Makefiles for init.c. For testfloat I've added the appropriate -march* options to use FPU instructions.

Conceivably we could make a different bsp for Fava vs Guava that would turn on the right GCC options when compiling for Fava. (Is it possible to do that via a bsp?)

mwachs5:
@aswaterman gave me a code snippet that would basically do this and some other setup that would be required if you had an FPU:

```
li t0, MSTATUS_FS
csrs mstatus, t0
csrr t1, mstatus
and t1, t1, t0
beqz t1, 1f
fsrr x0
1:
```

aswaterman:
That last instruction should be "fssr x0"--evidently I didn't try to assemble before sending that code.

The other option is to do what @scottj97 said and use ifdef __riscv_flen to prevent the code from being emitted when -march=RV32IMAC, rather than RV32IMAFC.  Should be OK either way.

scottj97:
I pushed another commit that now checks MISA to see if the F extension is present, and also initializes the rounding modes (which are undefined at reset).

aswaterman:
yep lgtm

PR #20, opened 2016-12-09
demo_gpio: Correct trap_entry handling to enable timer/ext interrupts

mwachs5:
Use the ASM trap_entry wrapper around handle_trap C function


mwachs5:
@rxia can we update the MCS to include this software? This demo is more interesting than it was.

richardxia:
> @rxia can we update the MCS to include this software? This demo is more interesting than it was.

Yep, I just uploaded it as v 1.0.2.

PR #23, opened 2016-12-14
Bump GNU toolchain

a0u:
This pulls in a significant suite of changes:
* Various ABI refinements and fixes
* Revised set of predefined macros
* Overhaul of compiler flags

### DIFF
--- a/bsp/env/freedom-e300-hifive1/entry.S
+++ b/bsp/env/freedom-e300-hifive1/entry.S
@@ -1,55 +1,47 @@
-// See LICENSE for License Details
+// See LICENSE for license details
+
 #ifndef ENTRY_S
 #define ENTRY_S
 
 #include "encoding.h"
-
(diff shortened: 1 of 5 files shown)


bsp/env/freedom-e300-hifive1/entry.S
```diff
@@ -1,55 +1,47 @@
-// See LICENSE for License Details
+// See LICENSE for license details
```
mwachs5: Do you think that we should make (perhaps as a different PR) these files common in a shared/common directory vs. simlinking?

a0u: Probably a common directory such as `bsp/env` (or perhaps one more appropriately named).  
The symlinking kept open the option of adding board-specific refinements later, but I'm no longer convinced that these are needed in the foreseeable future.

PR #24, opened 2016-12-14
Convert libc stubs to link-time wrappers

a0u:
Rather than abuse preprocessor macros (e.g., `-Dscanf=dhry_scanf`), this leverages the `--wrap=<symbol>` feature of GNU ld to substitute system functions at link time instead.

These wrappers are contained in their own static library (libwrap) for two reasons:
* Use `--start-group`/`--end-group` link options to prioritize our implementation but also coerce references in libc.a to resolve to our wrappers.
* Permit ld to eliminate dead code.


a0u:
I have tested dhrystone and the demos on the HiFive board.  It turns out that newlib's `printf()` works - it just causes a lot of unnecessary code to be statically linked due to `sprintf()`, which likely does not work since it calls `_malloc_r()`.

PR #26, opened 2016-12-14
Update README.md

mwachs5:
Add additional package dependency due to new versions of the tools


mwachs5:
This worked for me after adding the package.

PR #37, opened 2017-01-03
Fixes in the root makefile

NonerKao:
Each `cd` command in the targets **software**, **software_clean**, and
**dasm** is unnecessary.  Also, fixes from `make` to `$(MAKE)` help
the readability of build/dry-run output.


richardxia:
Thanks for submitting this!

PR #40, opened 2017-01-06
Use HTTPS when ever possible and speedup the first build

dns2utf8:
Hi all

I noticed that the readme used http for most links.
While cloning the repo, I noticed most of the submodules are http too. Since github redirects to https anyways I updated the module path.

Regards

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,6 +1,6 @@
 [submodule "openocd"]
 	path = openocd
-	url = http://github.com/sifive/openocd.git
+	url = https://github.com/sifive/openocd.git
 [submodule "riscv-gnu-toolchain"]
 	path = riscv-gnu-toolchain
-	url = http://github.com/riscv/riscv-gnu-toolchain.git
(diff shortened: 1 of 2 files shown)


aswaterman:
Does sh define $(nproc) or is that a bashism?

a0u:
> Does sh define $(nproc) or is that a bashism?

`nproc(1)` is a GNU-ism.  A more portable solution is:
```sh
_nproc() {
    { command -v nproc >/dev/null && nproc ; } ||
    getconf _NPROCESSORS_ONLN 2>/dev/null ||
    getconf NPROCESSORS_ONLN 2>/dev/null ||
    echo 1
}
```

dns2utf8:
`$( ... )` executes a sub-shell and inserts the output in the current commandline like:
```
` ... `
```

How portable should the manual be? I think while the function snippet works, it is too long to enter.
Maybe I should omit it or replace it with a good guess like `-j 4`?

dns2utf8:
Good points. I changed it to a text.
Should I rebase it before the merge?

Issue #34, opened 2016-12-28
Include led_fade demo that ships on hifive1 boards


tdaede:
demo_gpio.c is slightly different.

mwachs5:
PR #44 addresses this issue. PR #41 also addresses the issue, but 44 is the complete code.

mwachs5:
Fixed by #44

PR #41, opened 2017-01-07
Implement LED demo

dns2utf8:
I implemented something like the led demo (#34) with the PWM-block.

The cpu is way to fast for the led so I wrote `busy_loop` to slow down the color fade.

This code will produce a smooth fade: [see video](https://vimeo.com/198468534)

### DIFF
--- a/software/led_color_cycle/led_color_cycle.c
+++ b/software/led_color_cycle/led_color_cycle.c
@@ -1,0 +1,144 @@
+#include <stdio.h>
+#include "platform.h"
+
+// Do not set this to high, it may damage your led
+const uint8_t LED_MAX_BRIGHTNESS = 0x20;
+
+
+// Simple variables for LEDs, buttons, etc.
+volatile unsigned int* g_output_vals  = (unsigned int *) (GPIO_BASE_ADDR + GPIO_OUTPUT_VAL);
+volatile unsigned int* g_input_vals   = (unsigned int *) (GPIO_BASE_ADDR + GPIO_INPUT_VAL);
+volatile unsigned int* g_output_en    = (unsigned int *) (GPIO_BASE_ADDR + GPIO_OUTPUT_EN);
+volatile unsigned int* g_pullup_en    = (unsigned int *) (GPIO_BASE_ADDR + GPIO_PULLUP_EN);
+volatile unsigned int* g_input_en     = (unsigned int *) (GPIO_BASE_ADDR + GPIO_INPUT_EN);
+volatile unsigned int* g_iof_en       = (unsigned int *) (GPIO_BASE_ADDR + GPIO_IOF_EN);
+volatile unsigned int* g_iof_sel      = (unsigned int *) (GPIO_BASE_ADDR + GPIO_IOF_SEL);
+volatile unsigned int* g_out_xor      = (unsigned int *) (GPIO_BASE_ADDR + GPIO_OUTPUT_XOR);
+
+
+/**
+ * Wait a bit, TODO improve
+ */
+void busy_loop(uint32_t const duration) {
+  uint32_t j;
+  for (uint32_t i = duration; i > 0; --i) {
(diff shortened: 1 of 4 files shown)


mwachs5:
Nice! We do still plan to post the led_fade demo code that ships on the board, just working through the TODO list.

software/led_color_cycle/led_color_cycle.c
```diff
@@ -0,0 +1,144 @@
+#include <stdio.h>
+#include "platform.h"
+
+// Do not set this to high, it may damage your led
```
mwachs5: I don't think it will damage the LED, but it could damage your eyes!

software/led_color_cycle/led_color_cycle.c
```diff
@@ -0,0 +1,144 @@
...
+volatile unsigned int* g_input_en     = (unsigned int *) (GPIO_BASE_ADDR + GPIO_INPUT_EN);
+volatile unsigned int* g_iof_en       = (unsigned int *) (GPIO_BASE_ADDR + GPIO_IOF_EN);
+volatile unsigned int* g_iof_sel      = (unsigned int *) (GPIO_BASE_ADDR + GPIO_IOF_SEL);
+volatile unsigned int* g_out_xor      = (unsigned int *) (GPIO_BASE_ADDR + GPIO_OUTPUT_XOR);
+
+
+/**
+ * Wait a bit, TODO improve
+ */
+void busy_loop(uint32_t const duration) {
```
mwachs5: I like to use the RTC, by reading the mtime register in the CLINT to do this sort of delay. Or for more precise loops, you can use mcycle (but then you need to know your CPU freq).

software/led_color_cycle/led_color_cycle.c
dns2utf8: I agree, should I remove it?

dns2utf8:
I rewrote it to use the `mtime` register.

I am currently polling it to keep it simple. An example without interrupt handlers.

Currently I am wondering, what is the speed of `mtime`? Is it milliseconds?

mwachs5:
`mtime` runs at the RTC frequency, so 32.768 kHz.

software/led_color_cycle/led_color_cycle.c
```diff
@@ -0,0 +1,136 @@
...
+  printf("\n  hello led_color_cycle!\n");
+
+  // Enable LED output
+  * g_input_en    &= ~((0x1<< RED_LED_OFFSET) | (0x1<< GREEN_LED_OFFSET) | (0x1 << BLUE_LED_OFFSET)) ;
+  * g_output_en   |= ((0x1<< RED_LED_OFFSET) | (0x1<< GREEN_LED_OFFSET) | (0x1 << BLUE_LED_OFFSET)) ;
+  * g_output_vals |= ((0x1<< RED_LED_OFFSET) | (0x1<< GREEN_LED_OFFSET) | (0x1 << BLUE_LED_OFFSET)) ;
+
+  // Use PWM1 to generate output signal
+  *g_iof_en       |= ((0x1<< RED_LED_OFFSET) | (0x1<< GREEN_LED_OFFSET) | (0x1 << BLUE_LED_OFFSET)) ;
+  *g_iof_sel      |= IOF1_PWM1_MASK;
```
dns2utf8: I wonder if I enable too much with this selector.

mwachs5: Yes, the IOF mask also includes an output for PWM_CMP0, so it is controlling one more output than you really need.

mwachs5:
We are in the process of creating a separate repo for more demos and examples, to separate them from the SDK toolchain / BSP code. Will post here when that's set up!

dns2utf8:
Works for me 👍

palmer-dabbelt:
I'm closing this because it's two years old and has some conflicts.  Sorry we missed this!

PR #43, opened 2017-01-10
Implement demo UART application

dgrubb:
As a stepping stone for supporting serial based external devices I've created wrapper functions for using the UART peripheral (largely cribbed from the Arduino support libraries). A modest demo application is provided for validation.

Cheers.

### DIFF
--- a/software/demo_uart/Makefile
+++ b/software/demo_uart/Makefile
@@ -1,0 +1,8 @@
+TARGET = demo_uart
+C_SRCS += demo_uart.c
+C_SRCS += UART_driver.c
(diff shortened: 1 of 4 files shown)


mwachs5:
Thanks for your contribution. I agree that this sort of UART driver (for all the peripherals) is going to be useful. But since this code was taken from the Arduino library, it's under GPL, and we can't merge it to this repo (which is Apache 2.0) licensed.

dgrubb:
Opps, I'm a bit embarrassed for overlooking that. Having said that, it's heartening to see that SiFive's commitment to open-source is more than a statement and you're carefully considering the nuances of the different licenses.

PR #44, opened 2017-01-10
Add the led_fade demo which ships on HiFive1 boards

mwachs5:
This is the example program which ships on the HiFive1 boards.

### DIFF
--- a/software/led_fade/led_fade.c
+++ b/software/led_fade/led_fade.c
@@ -1,0 +1,160 @@
+// See LICENSE for license details.
+
+// This is the program which ships on the HiFive1
+// board, executing out of SPI Flash at 0x20400000.
+
+#include <stdint.h>
+#include "platform.h"
+
+static const char led_msg[] = "\a\n\r\n\r\
+55555555555555555555555555555555555555555555555\n\r\
+5555555 Are the LEDs Changing? [y/n]  555555555\n\r\
+55555555555555555555555555555555555555555555555\n\r";
+
+static const char sifive_msg[] = "\n\r\
+\n\r\
+                SIFIVE, INC.\n\r\
+\n\r\
+         5555555555555555555555555\n\r\
+        5555                   5555\n\r\
+       5555                     5555\n\r\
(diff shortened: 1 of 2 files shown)


software/led_fade/led_fade.c
```diff
@@ -0,0 +1,160 @@
...
+  PWM1_REG(PWM_CMP0)  = 0;
+
+  GPIO_REG(GPIO_IOF_SEL)    |= ( (1 << GREEN_LED_OFFSET) | (1 << BLUE_LED_OFFSET) | (1 << RED_LED_OFFSET));
+  GPIO_REG(GPIO_IOF_EN )    |= ( (1 << GREEN_LED_OFFSET) | (1 << BLUE_LED_OFFSET) | (1 << RED_LED_OFFSET));
+  GPIO_REG(GPIO_OUTPUT_XOR) &= ~( (1 << GREEN_LED_OFFSET) | (1 << BLUE_LED_OFFSET));
+  GPIO_REG(GPIO_OUTPUT_XOR) |= (1 << RED_LED_OFFSET);
+
+  while(1){
+    volatile uint64_t *  now = (volatile uint64_t*)(CLINT_BASE_ADDR + CLINT_MTIME);
+    volatile uint64_t then = *now + 100;
```
dns2utf8: Why is then volatile?

mwachs5: no good reason. Probably a vestige of an earlier version of the code.

dns2utf8:
Would you mind if I merged some of my code into yours? Since it does basically the same it may be preferable to keep just one.

mwachs5:
I think it is best to keep just one. I think people should have the right to know what is actually shipped on their board, so I would prefer to keep the led_fade demo as-as.

dns2utf8:
While I agree with knowing what is inside the box, I think this code will be copied many times because it is **the** example inside the sdk for working with the LED.

A compromise would be to link to the current commit from the readme and cleanup the example.
What do you think?

PR #46, opened 2017-01-25
Various SDK Cleanup Items

mwachs5:
Clean up various items in the SDK:

* Be more consistent in the OpenOCD Configs between Arty and HiFive1
* Use latest riscv-gnu-toolchain to properly support 'C' extension in GDB
* Use register macros in demo_gpio, and add a bit-banging speed demo


mwachs5:
@aswaterman , @a0u any known reason that I can't bump to the lastest riscv-gnu-toolchain? It seems to work.

PR #48, opened 2017-01-27
Update README.md

mwachs5:
Add details about how to keep the repo up to date


README.md
```diff
@@ -43,6 +43,23 @@ To see additional options:
...
+git pull origin master
+git submodule update --init --recursive
+```
+
+If you would like to recompile the entire toolchain after performing the above:
+
+```
+rm -rf toolchain
+rm -rf work
+make tools
```
richardxia: Should there be a `make clean` or `make clean-tools` target to do the `rm -rf` commands?

mwachs5: Turns out there was actually already an 'uninstall' target that pretty much did that. I cleaned it up and used that in the README instead.

PR #53, opened 2017-03-01
Add Performance Counter Demo

mwachs5:
This adds a simple demo to show how to use the low level performance counters for profiling code.

This also sets NO_INIT for the `led_fade` demo, which makes it compile properly as it is actually shipped on the board.

### DIFF
--- a/software/performance_counters/performance_counters.c
+++ b/software/performance_counters/performance_counters.c
@@ -1,0 +1,111 @@
+// See LICENSE for license details.
+
+// This demo shows how to use basic
+// RISC-V profiling counters, mcycle
+// (counts the number of processor cycles)
+// and minstret (counts the number of retired instructions). 
+// Note that both are writable as well.
+
+#include <stdio.h>
+
+// The CSR encodings are in this header.
+#include "encoding.h"
+
+// The mcycle counter is 64-bit counter, but since
+// Freedom E platforms use RV32, we must access it as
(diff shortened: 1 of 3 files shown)


software/performance_counters/performance_counters.c
```diff
@@ -0,0 +1,111 @@
...
+#include "encoding.h"
+
+// The mcycle counter is 64-bit counter, but since
+// Freedom E platforms use RV32, we must access it as
+// 2 32-bit registers. At 256MHz, the lower bits will
+// roll over approx. every 5 seconds, so we check for
+// rollover with this routine as suggested by the
+// RISC-V Priviledged Architecture Specification.
+
+#define rdmcycle(x)  {				       \
```
richardxia: Does this need to be defined as a C macro? Since this is a C file, you can just define it as a normal function.

mwachs5: well, it should be inlined, etc.... we don't really want it taking extra cycles to return stuff. 

But agreed that this can/should be moved to a shared library.

richardxia: I believe -O2 will automatically inline small functions. The manual says

> Integrate functions into their callers when their body is smaller than expected function call code (so overall size of program gets smaller). The compiler heuristically decides which functions are simple enough to be worth integrating in this way. This inlining applies to all functions, even those not declared inline.

But there's also a directive we can add to force it to always inline. Anyway, like I said, not a blocking concern.

PR #59, opened 2017-05-17
Update Makefile and README with how to use precompiled binaries.

mwachs5:
I haven't tested the flow or instructions yet, but let me know if this seems like a reasonable way to handle it in the Makefile


dbarbi1:
Method looks fine, though I dont like the variable names with PREFIX. Can we use RISCV_PATH or RISCV_DIR?

PR #60, opened 2017-05-19
FreeRTOS

dbarbi1:
this PR changes the prints in _init to %ld to fix the compilation warning. Everything else is new to support FreeRTOS


mwachs5:
Doesn't the %ld depend on whether you are compiling for 64 or 32 bit? @palmer-dabbelt?

palmer-dabbelt:
On RV32 long is 32 bits, and on RV64 long is 64 bits.  You might want something like PRId64.

dbarbi1:
this was generating a warning for me in the FreeRTOS builds. Now checking the standard sdk demos, i'm not seeing the warning when using the %d init file....

Is it possible my build arguments are missing something or incorrect?

dbarbi1:
get_cpu_freq returns an unsigned long though, shouldn't we be using %ld?

palmer-dabbelt:
If you're trying to print an unsigned long then you should use "%lu".

dbarbi1:
Closing request. We are going to put FreeRTOS into it's own repo outside of FESDK

PR #62, opened 2017-06-01
Dead script

edwardcwang:
That script seems to be gone?

@mwachs5


palmer-dabbelt:
Yes, this is defunct now that we use GDB to upload instead of OpenOCD.

PR #69, opened 2017-07-26
Vector

dbarbi1:
This PR adds a vectored interrupt example for E31/E51 Arty targets.


bsp/env/ventry.S
```diff
@@ -6,123 +6,54 @@
...
-  STORE x25, 25*REGBYTES(sp)
-  STORE x26, 26*REGBYTES(sp)
-  STORE x27, 27*REGBYTES(sp)
-  STORE x28, 28*REGBYTES(sp)
-  STORE x29, 29*REGBYTES(sp)
-  STORE x30, 30*REGBYTES(sp)
-  STORE x31, 31*REGBYTES(sp)
+  addi sp, sp, -18*REGBYTES
+
+  STORE x1,  1*REGBYTES(sp)
```
aswaterman: start at ```0*REGBYTES``` instead of 1?

bsp/env/ventry.S
```diff
@@ -6,123 +6,54 @@
...
+  STORE x7,  4*REGBYTES(sp)
+  STORE x10, 5*REGBYTES(sp)
+  STORE x11, 6*REGBYTES(sp)
+  STORE x12, 7*REGBYTES(sp)
+  STORE x13, 8*REGBYTES(sp)
+  STORE x14, 9*REGBYTES(sp)
+  STORE x15, 10*REGBYTES(sp)
+  STORE x16, 11*REGBYTES(sp)
+  STORE x17, 12*REGBYTES(sp)
+  STORE x18, 13*REGBYTES(sp)
```
aswaterman: Don't need to save/restore ```x18```

dbarbi1: removed, thanks

bsp/env/ventry.S
dbarbi1: the standard entry.S starts at 1*REGBYTES. does it need to be changed too?

https://github.com/sifive/freedom-e-sdk/blob/master/bsp/env/entry.S#L15

aswaterman: I'd leave that one alone.  There, the reasoning was different... note the ```mv a2, sp```.  It's passing the array of registers as the third argument ```handle_trap```.  Presumably, the C code would like it so ```regs[1]``` corresponds to ```x1```.

aswaterman: Also, technically, the stack is supposed to have 16-byte alignment.  So, using ```32*REGBYTES``` in the standard entry.S, and ```16*REGBYTES``` in the optimized one, does just the right thing.

dbarbi1: ok. Note that I kept the code the same for the synchronous trap vector and the C code handler:
https://github.com/sifive/freedom-e-sdk/blob/vector/bsp/env/ventry.S#L131
https://github.com/sifive/freedom-e-sdk/blob/vector/software/vectored_interrupts/vectored_interrupts.c#L79

Should I change the vector entry to be the same as the others, and just perform the CSR reads in the C code handler?

dbarbi1: I went ahead made the changes, including the synchronous trap handler now reading _mcause_ in the C handler.

aswaterman: Doing it in assembly code makes it consistent with the other example, but doing it in C code seems cleaner... it's a toss-up, so whatever you think is best

PR #80, opened 2017-11-21
E31 Dhrystone Linker script

palmer-dabbelt:
This gets Dhrystone in an E31 for me, it might require some twiddling for the E51.


palmer-dabbelt:
https://github.com/sifive/freedom-e-sdk/pull/81

PR #81, opened 2017-11-21
Dhrystone arty

dbarbi1:
Update Dhrystone for coreplexip arty builds. 

`make software PROGRAM=dhrystone BOARD=coreplexip-eX1-arty LINK_TARGET=dhrystone`


palmer-dabbelt:
I think they're the same

PR #82, opened 2017-11-21
Make space for e51 dhrystone

dbarbi1:
reduce stack size so that E51 dhrystone fits in 16k DTIM


dbarbi1:
sorry, i thought this was already in the previous PR

PR #84, opened 2018-01-03
Add _FUNC aliases for newlib-2.5.0

palmer-dabbelt:
I'm not sure what the policy actaully is here, but some newlib functions
now have an extra underscore before them.  This defines a bunch of
aliases and some more wrappers to link against the correct newlib
functions for 2.5.0 but maintain compatibility with 2.4.0.

### DIFF
--- a/bsp/libwrap/libwrap.mk
+++ b/bsp/libwrap/libwrap.mk
@@ -42,6 +42,7 @@ LIBWRAP := libwrap.a
 LINK_DEPS += $(LIBWRAP)
 
 LDFLAGS += $(foreach s,$(LIBWRAP_SYMS),-Wl,--wrap=$(s))
+LDFLAGS += $(foreach s,$(LIBWRAP_SYMS),-Wl,--wrap=_$(s))
 LDFLAGS += -L. -Wl,--start-group -lwrap -lc -Wl,--end-group
 
 CLEAN_OBJS += $(LIBWRAP_OBJS)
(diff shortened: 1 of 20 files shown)


dbarbi1:
This fixes the issues I was experiencing. Seems silly for function names to change like this between newlib version. A quick google didn't turn anything up... but i dont know where to look for these things

ilg-ul:
> I'm not sure what the policy actaully is here, but some newlib functions now have an extra underscore before them.

Things are more complicated then you think, I tried to advocate for this several month ago.

The normal use case is to have the application define the functions **with** underscore. The way your initial versions used functions without undercore was not ok.

As long as you do not affect newlib, it is not my business, you can do whatever you want, but I recommend you to update your code to use the correct names, with underscore, and do not play tricks with renaming macros, you may trigger unwanted side effects.

jim-wilson:
This was changed by a local newlib/libgloss patch.
    https://github.com/riscv/riscv-newlib/pull/21/files

As Liviu mentioned, it was always wrong for riscv libgloss to be defining functions like kill.  newlib provides a kill function that calls _kill_r which provides re-entrant support, and then calls _kill, which is provided by target dependent code.  If you redefine kill, you break reentrant support, and possibly cause other problems.  The current libgloss riscv support looks correct.  The older libgloss riscv support was broken.

ilg-ul:
Thank you, Jim. 

(Finally someone who understood the problem with the old toolchain...)

PR #89, opened 2018-03-29
bsp:devices:spi.h: fix sckmode fields value

trabucayre:
According to : 
- SiFive FE310-G000 Manual;
- src/main/scala/devices/spi/TLSPI.scala on github sifive-blocks
- and control with logic analyzer.
CPHA is the bit 0 of sckmod register (ie value 0 or 1) and CPOL is the bit 1 of sckmod register (ie value 0 or 2).
This patch correct values for these to define

### DIFF
--- a/bsp/include/sifive/devices/spi.h
+++ b/bsp/include/sifive/devices/spi.h
@@ -30,8 +30,8 @@
 
 /* Fields */
 
-#define SPI_SCK_POL             0x1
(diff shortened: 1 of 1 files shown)


trabucayre:
Hello,
It seems to have a problem with your travis. A missing fire for rake visibly

palmer-dabbelt:
We don't have travis for this repo so I turned it off.

PR #90, opened 2018-04-03
The HiFive1 has a 128Mb flash

palmer-dabbelt:
The website says there's a 128Mb flash on the HiFive1, but the linker scripts say it's 512MB.

### DIFF
--- a/bsp/env/coreplexip-e21-arty/dhrystone.lds
+++ b/bsp/env/coreplexip-e21-arty/dhrystone.lds
@@ -1,0 +1,1 @@
+../coreplexip-e31-arty/dhrystone.lds
--- a/bsp/env/coreplexip-e21-arty/flash.lds
+++ b/bsp/env/coreplexip-e21-arty/flash.lds
@@ -1,0 +1,1 @@
+../coreplexip-e31-arty/flash.lds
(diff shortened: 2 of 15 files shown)


mwachs5:
Half of that flash is actually used for the FPGA. So it should only be 8M if you are starting at 0x20400000.

palmer-dabbelt:
The HiFive1 doesn't have an FPGA... :)

mwachs5:
True. 
But this is still using the first 4MB then for the "safe bootloader" code. That's why it's starting at 0x20400000 instead of 0x2000_0000. So I guess the size is most accurately 12MB.

ilg-ul:
>  first 4MB then for the "safe bootloader" code

a very useful feature, given how easy it is to brick the board. (while trying to do active development on the board, abruptly terminating openocd, like when canceling a debug session, most of the time leaves the board unusable, with openocd no longer able to connect to the board; the only solution I found was to bring the board in bootloader mode, and restart the debug session).

so, although techincally the Arduino bootloader is not mandatory for bare-metal development, it is highly recommended to preserve the bootloader, and have the application loader scripts skip the first 4 MiB, leaving only 12 MiB for the application.

palmer-dabbelt:
Thanks, I updated it to 12MB

PR #96, opened 2018-07-03
Clic vectored

dbarbi1:
This PR adds support for the E2 Series Arty FPGA images along with a vectored interrupt example application.


dbarbi1:
@palmer-dabbelt i think this is ready to merge if you are ok with it. Note that it breaks clic_interrupts example and also requires a toolchain bump.

PR #107, opened 2018-11-03
Use the appropriate toolchain (32 vs 64)

borancar:
I had issues using my own compiled toolchain from https://github.com/sifive/freedom dependency https://github.com/freechipsproject/rocket-chip.

Basically anyone who has followed https://github.com/freechipsproject/rocket-chip will have 2 toolchains installed, riscv32-unknown-elf- and riscv64-unknown-elf-, which will cause problems with using the wrong compiler here (and errors ELFCLASS64 vs ELFCLASS32 at linking stage), so let's just apply the right name to the toolchain.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -114,11 +114,11 @@ toolchain_prefix := $(toolchain_builddir)/prefix
 
 RISCV_PATH ?= $(toolchain_prefix)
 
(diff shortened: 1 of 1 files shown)


palmer-dabbelt:
I'm closing this in favor of #110 .  Since `riscv64-unknown-elf` can generate 32-bit binaries it's not correct to switch just on XLEN, but instead the user needs to decide which toolchain they want to use.

PR #110, opened 2018-12-12
Allow users to set a CROSS_COMPLIE argument

palmer-dabbelt:
This allows users to select a different cross complier than the default
'riscv64-unknown-elf'.

Signed-off-by: Palmer Dabbelt <palmer@dabbelt.com>

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -2,6 +2,9 @@
 # Configuration
 #############################################################
 
+# Allow users to select a different cross compiler.
+CROSS_COMPILE ?= riscv64-unknown-elf
+
 # Allows users to create Makefile.local or ../Makefile.project with
 # configuration variables, so they don't have to be set on the command-line
 # every time.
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -2,6 +2,9 @@
 # Configuration
 #############################################################
 
+# Allow users to select a different cross compiler.
+CROSS_COMPILE ?= riscv64-unknown-elf-
```
nategraff-sifive: It's probably more ergonomic to not expect the user to include the trailing '-'

palmer-dabbelt: Linux makes you stick the "-" in there, which is where I stole the variable from.  I'm OK changing it if you write the code :)

nategraff-sifive: If Linux expects it I'm fine with the principle of least surprise.

palmer-dabbelt: OK, I just did it :)

PR #115, opened 2018-12-14
Add MEE BSP for S76 CoreIP

nategraff-sifive:
Test against RTL simulation before merge. Also double-check port protocols in DTS against RTL deliverable.

### DIFF
--- a/bsp/coreip-s76/mee.lds
+++ b/bsp/coreip-s76/mee.lds
@@ -1,0 +1,189 @@
+OUTPUT_ARCH("riscv")
+
+ENTRY(_enter)
+
+MEMORY
+{
+	ram (wxa!ri) : ORIGIN = 0x20000000, LENGTH = 0x4000000
+}
+
+PHDRS
+{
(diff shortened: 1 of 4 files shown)


bsp/coreip-s76/mee.lds
```diff
@@ -0,0 +1,189 @@
...
+		PROVIDE( mee_segment_data_target_start = . );
+	} >ram AT>ram :ram_init
+
+
+	.data 		:
+	{
+		*(.data .data.*)
+		*(.gnu.linkonce.d.*)
+		. = ALIGN(8);
+		PROVIDE( __global_pointer$ = . + 0x800 );
```
bsousi5: I seem to recall, but does 0x800 always for all cores? May want to check with palmer

palmer-dabbelt: This actually depends on the benchmark more than anything else.  IIRC Jim had a better scheme, but we can put it in the linker script generator later.

PR #116, opened 2018-12-14
Fix COREIP_HEX_WIDTH

palmer-dabbelt:
I must have been too tired last night...


bsp/coreip-e31/settings.mk
```diff
@@ -2,3 +2,4 @@
 
 RISCV_ARCH=rv32imac
 RISCV_ABI=ilp32
+COREIP_MEM_WIDTH=32
```
nategraff-sifive: Should this be set for coreip-s51 as well?

PR #119, opened 2018-12-14
Add script for updating MEE BSPs

nategraff-sifive:


nategraff-sifive:
@palmer-dabbelt I put this script under `bsp/`, maybe it should get moved to `scripts/`.

PR #122, opened 2018-12-18
Example of Software interrupt

bsousi5:
A simple example on how to use mee interrupts APIs.
The mee interrupts drivers require updating DTS files, mee.h headers and an update of the freedom-mee submodule.


nategraff-sifive:
Can we also add an example to demonstrate how to create an exception handler?

PR #123, opened 2018-12-18
Pass AR through when building MEE-based examples

palmer-dabbelt:

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -231,6 +231,7 @@ $(PROGRAM_ELF): \
 		$(MEE_BSP_PATH)/install/lib/libmee-gloss.a \
 		$(MEE_BSP_PATH)/mee.lds
 	$(MAKE) -C $(dir $@) $(notdir $@) \
+		AR=$(RISCV_AR) \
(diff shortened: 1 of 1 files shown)


sifivekevin:
Palmer,  This change isn't working for me.  I don't think they are being passed through as expected. The generated makefile is still trying to use `ar`.  I see in `freedom-mee/Makefile.in` that AR is explicitly as `AR = ar` (while `CC = @CC@`).    I tried `AR = @AR@`, but to no avail.  It works if I set `AR = riscv64-unknown-elf-ar` in Makefile.in (then rebuild the bsp).  I'm continuing to work on how to fix it, but it may be more obvious to you as I know nothing about these makefile generation tools.

sifivekevin:
Actually the change you made is probably working fine (being passed down to the software makefile), but the failure looks like it is happening during the bsp=mee build (before the change you made passes it down).

palmer-dabbelt:
https://github.com/sifive/freedom-metal/pull/22
https://github.com/sifive/freedom-e-sdk/pull/125

PR #125, opened 2018-12-19
Use cross-AR, round 2

palmer-dabbelt:
This uses cross-AR for both the MEE build and the example build.


palmer-dabbelt:
Depends on https://github.com/sifive/freedom-metal/pull/22

PR #129, opened 2019-01-03
Partial fixes to get coremark and dhrystone building again.

jim-wilson:
Still some mee issues to resolve, such as using the new timing functions that
haven't been added yet, and replacing the platform.h and encoding.h headers.
Also, other software dirs will need similar changes.

### DIFF
--- a/software/coremark/core_portme.c
+++ b/software/coremark/core_portme.c
@@ -1,8 +1,8 @@
 #include <stdio.h>
 #include <stdlib.h>
 #include "coremark.h"
-#include "platform.h"
-#include "encoding.h"
+//#include "platform.h"
+//#include "encoding.h"
 
 #if VALIDATION_RUN
 	volatile ee_s32 seed1_volatile=0x3415;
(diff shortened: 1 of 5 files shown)


software/coremark/core_portme.c
```diff
@@ -1,8 +1,8 @@
 #include <stdio.h>
 #include <stdlib.h>
 #include "coremark.h"
-#include "platform.h"
-#include "encoding.h"
+//#include "platform.h"
+//#include "encoding.h"
```
nategraff-sifive: If we still want coremark to build with the legacy BSP, these should probably still be included somehow.

jim-wilson: I have no idea what is going on with freedom-e-sdk.  I just see breaking changes being added without any apparent plan, and on master instead of on a branch.  So I have no idea how to fix this.  But my patch does at least let me build some broken stuff using BSP=mee which is apparently the intent of the breaking changes.

brucehoult: "If we still want coremark to build with the legacy BSP"

I don't know. Do we? Long term, shouldn't MEE be replacing the legacy BSP?

Now, when MEE doesn't work, certainly. Which makes it disappointing that WIP stuff for MEE is getting checked in on master branch and breaking it. Normal people actually use freedom-e-sdk. Why isn't this work happening on a branch? Telling people "Check out revision baeeb8f from September" (or something) isn't the right way to do it,. Master should simply work. Always.

@palmer-dabbelt

nategraff-sifive: Is legacy broken? All the MEE changes should still be isolated behind `BSP=mee`. Either way you're right, a development branch is a good idea.

I made the original comment here because this change would be mutually exclusive between legacy and MEE.

PR #148, opened 2019-01-15
Update readme

nategraff-sifive:
- Note that the repository is under construction
- Explain what Freedom Metal is
- Clarify what the arguments to `make` mean
- Explicitly list which examples are for which BSP
- Update `make help`


README.md
```diff
@@ -82,47 +142,72 @@ export RISCV_PATH=/my/desired/location/riscv64-unknown-elf-gcc-<date>-<version>
...
 ```
 
 ### Using the Tools ###
 
+To ease the transition for consumers of the Legacy SDK, we've namespaced
+the new Freedom Metal examples behind the environment variable `BSP`. To use the
+Freedom Metal examples, you may `export BSP=mee` in your shell environment, or simply
+pass `BSP=mee` as an argument to your `make` invocations as you see in the following
+example commands.
+
```
bsousi5: We also need a section on BSP, describing its contents and what is relevant for mee vs lagacy.
For example mee, does not use env, drivers, include/sifive and liberal folders. But only utilize the target dts, header and linker files in each of the support target directories under bsp.

nategraff-sifive: I added a description of the BSP file locations to the **Contents** section.

PR #151, opened 2019-01-16
Standalone project export

nategraff-sifive:
Splits out board init, toolchain setup, and program build into `scripts/standalone.mk` which can be included in the top-level Makefile to allow in-SDK builds or be exported and used on its own.

Creates a `standalone` Make target which copies an example, the built Freedom Metal library, and the standalone Makefile into an external project.


README.md
```diff
@@ -235,6 +235,17 @@ make BSP=mee [PROGRAM=hello] [BOARD=sifive-hifive1] debug
 make BSP=mee [PROGRAM=hello] [BOARD=sifive-hifive1] clean
 ```
 
+#### Create a Standalone Project ####
+
+You can export a program to a standalone project directory using the `standalone`
+target. The resulting project will be locked to a specific target `BOARD`.
+
```
bsousi5: Can explicitly state Standalone only support/applicable for BSP=mee

sifivekevin:
Works great on Linux and almost works on Windows/msys.  For some reason, this line from the makefile:

BSP_FILES_TO_COPY = $(filter-out $(BSP_DIR)/build,$(wildcard $(BSP_DIR)/*)) $(BSP_DIR)/install

ends up with two references to the $(BSP_DIR)/install directory and that breaks 'cp' in msys.  If I remove the second instance by changing the line to:

BSP_FILES_TO_COPY = $(filter-out $(BSP_DIR)/build,$(wildcard $(BSP_DIR)/*)) ~~$(BSP_DIR)/install~~

Then is works great on Linux and Windows/msys too.

sifivekevin:
I need to retract my "works great on Windows" statement.  The makefile change is still correct, but I've run into some build issues (unrelated to this PR) where the msys/mingw make is failing.  This PR sill looks good (with the suggested makefile correction in my previous comment).

sifivekevin:
I think [msys](http://www.mingw.org/wiki/MSYS) is just too old.  'make' is from 2006. It appears to be unmaintained at this point.  The good news is I did get the full build to work using [msys2](http://www.msys2.org/) (a fork of msys that is much newer) with the appropriate dependencies installed and after running autoreconf (which may or may not have been necessary).  The bad news is that including this environment will add 100's of megabytes to the installer unless I can figure out how to delete all the base stuff we don't need, if even possible.

cgsfv:
In the current Windows Freedom Studio package, we include a few linux-tools under the build-tools folder (make, echo, mkdir, rm, ...). These might originate from GNU MCU, as this is what requires these tools. Maybe we could follow the same approach, and at the same time try to minimize the number of special linux-tools used in the Makefiles?

cgsfv:
A way to limit the number of linux-tools used, could be to write it using python instead, as this is something that we anyway probably needs to ship along with Freedom Studio (gdb and qemu requires python)

cgsfv:
Actually, except for make, the other tools originate from this multi-tool, which supports many standard linux-utils on windows:

BusyBox v1.28.0.git (2018-01-03 19:20:27 UTC) multi-call binary
BusyBox is copyrighted by many authors between 1998-2015.
Licensed under GPLv2. See source distribution for detailed
copyright notices.

Usage: busybox [function [arguments]...]
   or: busybox --list
   or: busybox --install [DIR]
   or: function [arguments]...

        BusyBox is a multi-call binary that combines many common Unix
        utilities into a single executable.  The shell in this build
        is configured to run built-in utilities without $PATH search.
        You don't need to install a link to busybox for each utility.
        To run external program, use full path (/sbin/ip instead of ip).

Currently defined functions:
        [, [[, ar, arch, ash, awk, base64, basename, bash, bunzip2, busybox, bzcat, bzip2, cal, cat, chmod, cksum,
        clear, cmp, comm, cp, cpio, cut, date, dc, dd, df, diff, dirname, dos2unix, dpkg-deb, du, echo, ed, egrep, env,
        expand, expr, factor, false, fgrep, find, fold, ftpget, ftpput, getopt, grep, groups, gunzip, gzip, hd, head,
        hexdump, id, ipcalc, kill, killall, less, link, ln, logname, ls, lzcat, lzma, lzop, lzopcat, man, md5sum,
        mkdir, mktemp, mv, nc, nl, od, paste, patch, pgrep, pidof, printenv, printf, ps, pwd, rev, rm, rmdir, rpm,
        rpm2cpio, sed, seq, sh, sha1sum, sha256sum, sha3sum, sha512sum, shuf, sleep, sort, split, stat, strings, sum,
        tac, tail, tar, tee, test, timeout, touch, tr, true, truncate, uname, uncompress, unexpand, uniq, unix2dos,
        unlink, unlzma, unlzop, unxz, unzip, usleep, uudecode, uuencode, vi, watch, wc, wget, which, whoami, whois,
        xargs, xxd, xz, xzcat, yes, zcat

sifivekevin:
MAybe ignore my previous comment about the duplicate "install" causing problems with the 'cp' command.  I think that may be more due to the really old msys implementation.  With msys2 I had to undo my Makefile edit in order to get the install directory copied.  So no changes requested there.

nategraff-sifive:
My last force-push included a more elegant solution to the behavior anyway. The real problem is that `$(wildcard ...)` is evaluated at the beginning, instead of during the execution of the rule, so if the install directory wasn't created already it wouldn't get picked up.

Instead, I'm running `$(shell ls ...)` during the execution of the rule, which picks up the install directory properly.

sifivekevin:
Perfect.  That was very likely what I was seeing.
—
*Kevin Mills*
Mobile: 503-267-9505 <15032679505>
http://www.sifive.com
kevin.mills@sifive.com

cgsfv:
Just tried to use a standalone project from within Freedom Studio on MacOS. After having added the make target 'all' to the root Makefile it worked just fine. Freedom Studio uses 'make all' and 'make clean' as the 2 major targets for building and cleaning. So could you add that 'all' target?

bsousi5:
Nice

PR #152, opened 2019-01-23
Disable maintainer mode in MEE build

nategraff-sifive:
To support users without automake/autoconf/m4, disable maintainer mode
during the Freedom Metal build to make sure that the autoconf scripts
don't try to run automatically.


sifivekevin:
Nate,  I may have been premature here.  I'm seeing the "autotools" mee build happening on all three of Ubuntu, Windows/msys, and Mac OS when I do a fresh clone of the repository ('master') then a standalone make.  Subsequent builds behave as expected.  I verified the '-disable-maintainer-mode' switch was present.

PR #161, opened 2019-01-29
Fix list-programs on MacOS

sifivekevin:
Remove '-r' from sed command (not required and does not work on MacOS).  Tested on Linux/Msys/MacOS.
Remove 'echo Done', no longer needed. I found a better way.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -119,10 +119,9 @@ list-boards:
 
 # MEE programs are any submodules in the software folder
 list-programs:
-	@echo program-list: $(shell grep -o '= software/.*$$' .gitmodules | sed -r 's/.*\///')
+	@echo program-list: $(shell grep -o '= software/.*$$' .gitmodules | sed 's/.*\///')
 
 list-options: list-programs list-boards
-	@echo done
 
 endif
 


Makefile
```diff
@@ -119,10 +119,9 @@ list-boards:
 
 # MEE programs are any submodules in the software folder
 list-programs:
-	@echo program-list: $(shell grep -o '= software/.*$$' .gitmodules | sed -r 's/.*\///')
+	@echo program-list: $(shell grep -o '= software/.*$$' .gitmodules | sed 's/.*\///')
 
 list-options: list-programs list-boards
```
cgsfv: What does list-options do - run both list-programs and list-boards?

sifivekevin:
Yes. Runs both other targets.  Freedom Studio runs 'list-options' and
parses the output for use in the New Project Wizard UI.  I do that instead
of running the make twice, once for each other target.

The other two targets are not strictly required. We could combine them into
a single target.  Maybe a better name too, like 'fs-wizard-options'. That
would help discourage unintended edits that would break FS's ability to
parse.

PR #165, opened 2019-01-30
Add TARGET as a synonym for BOARD

nategraff-sifive:


sifivekevin:
Nate,

The local-interrupt example fails to build when I use TARGET=X, but works when I use BOARD=X, for any X

sifivekevin:
I pulled the latest changes about 3 minutes ago and it still fails.
Another pull pulled nothing new just now.

sifivekevin:
Wait. Let me do a little more testing.

sifivekevin:
I take that back.  It does appear to be working with the latest change.  I
saw a different error, unrelated.

sifivekevin:
Nate, I found another problem.   The "list-targets" is treated as an empty recipe and not a fall-through to the list-boards recipe.  This breaks the list-options target.  If I change list-targets to be:

list-targets: **list-boards**
list-boards:
    @echo bsp-list: $(sort $(filter-out $(EXCLUDE_TARGET_DIRS),$(notdir $(wildcard bsp/*))))

sifivekevin:
Nate,  I'm also fine just eliminating the list-boards target (in favor of the list-targets target).  I don't think we need to worry about backward compatibility here since FS is the only known consumer of the target (and only in the last couple of days).

nategraff-sifive:
@sifivekevin I removed list-boards

sifivekevin:
Since this PR is not "sssigned" to anyone, does that mean I can do the merge and close it?

bsousi5:
Please go ahead an merge, if you need to use the change.
I just took a look, and it is fine with me.

Bunnaroath

PR #168, opened 2019-01-31
E24 arty support

bsousi5:


bsousi5:
@ahuntington Please take a look at DTS file as I modified/added some new fields

bsousi5:
@ahuntington Sure.
The DTS nodes on LEDs, Buttons, Switches and Clocks are hand added to allow clocking the serial port, since these information are not currently available in the generated DTS.
In addition, I added sifive,numintbits = <4>; to the clic node, as that has not been generated.

PR #170, opened 2019-02-12
Add hw-exec-breakpoint support in dts

bsousi5:


bsousi5:
@sifivekevin Let me know if you need some sort of header files to be generated.

PR #171, opened 2019-02-12
Enable Freedom Metal builds in standalone projects

nategraff-sifive:
Adds the `INCLUDE_METAL_SOURCES` option to the standalone target to build Freedom Metal within standalone projects instead of including only the built archive.

For example:

`make BSP=metal TARGET=coreip-e31 PROGRAM=hello STANDALONE_DEST=../hello INCLUDE_METAL_SOURCES=1 standalone`


Makefile
```diff
@@ -213,8 +171,28 @@ $(STANDALONE_DEST):
 $(STANDALONE_DEST)/%:
 	mkdir -p $@
 
-# We have to use $$(shell ls ...) in this recipe instead of $$(wildcard) so that we
-# pick up $$(BSP_DIR)/install
+ifneq ($(INCLUDE_METAL_SOURCES),)
```
bsousi5: what/where set INCLUDE_METAL_SOURCES?
I assume it is set as part of the make cmd line, if so can you also update the Makefile help

scripts/libmetal.mk
```diff
@@ -0,0 +1,48 @@
...
+		--with-machine-header=$(abspath $(MEE_HEADER)) \
+		--with-machine-ldscript=$(abspath $(MEE_LDSCRIPT)) \
+		--with-builtin-libgloss
+	touch -c $@
+
+$(BSP_DIR)/install/stamp: $(BSP_DIR)/build/Makefile
+	$(MAKE) -C $(abspath $(BSP_DIR)/build) install
+	date > $@
+
+$(BSP_DIR)/install/lib/libriscv%.a: $(BSP_DIR)/install/stamp ;@:
```
bsousi5: Not familiar with ";@"?

nategraff-sifive: That was in the original makefile too. What are the odds it was a typo?

nategraff-sifive: Nope, that seems to be necessary.

nategraff-sifive:
I've added the requested documentation and removed all .git* files from the copied Metal sources.

scripts/libmetal.mk
bsousi5: Sounds good.

PR #181, opened 2019-02-23
A pair of small fixes

palmer-dabbelt:
Depends on https://github.com/sifive/freedom-metal/pull/57


nategraff-sifive:
I broke out the ASFLAGS commit into its own PR.

I think software-elf and software-hex is less relevant given recent changes.

HTIF support should get merged into Freedom Metal master before it gets pulled in.

nategraff-sifive:
Closing but leaving the branch alone just in case.

PR #187, opened 2019-03-04
Coreplex welcome port to metal

dconn-sifive:
Ported coreplexip_welcome to metal using the name coreplexip-welcome.


software/coreip-welcome/coreip-welcome.c
```diff
@@ -0,0 +1,156 @@
...
+    printf("             55555           55555\n");
+    printf("               55555       55555\n");
+    printf("                 55555   55555\n");
+    printf("                   555555555\n");
+    printf("                     55555\n");
+    printf("                       5\n");
+    printf("\n");
+
+    printf("\n");
+    printf("Welcome to the E31/E51 Core IP FPGA Evaluation Kit!\n");
```
bsousi5: Best to start avoid specific coreIP type, since the number of variant now increases.
May be something like "printf("Welcome to the Standard Core IP FPGA Evaluation Kit!\n");"
Maybe freedom-metal should have an API to return this string

dconn-sifive: Ok I fixed the welcome banner message and updated the submodule - thanks

PR #191, opened 2019-03-06
Make update script independent of bash

bsousi5:


bsousi5:
@nategraff-sifive 
@nategraff-sifive 
> Can you split out the "Make update script independent of bash" into a separate PR?

Done

PR #195, opened 2019-03-07
Remove legacy BSPs and Software

nategraff-sifive:


sifivekevin:
The `BSP=metal` is no longer necessary, but ignored if present?

nategraff-sifive:
@sifivekevin Yes, it's silently ignored.

dbarbi1:
Can we put a note in the docs that the legacy Freedom-E-SDK can be found on a specific branch which will be preserved? I believe it was the 1_0 branch?

PR #196, opened 2019-03-07
Update Metal with GPIO and Cache APIs

nategraff-sifive:
Update to the latest Metal library which includes public APIs for GPIO output and cache controllers.


nategraff-sifive:
@bsousi5 The s51-arty works with both hello and local-interrupt.

PR #201, opened 2019-03-08
Remove information in README

nategraff-sifive:
Use the Sphinx documentation as the source of truth so we don't have to
maintain two copies.


nategraff-sifive:
@dbarbi1 You okay with this? I want your feedback before I merge.

PR #202, opened 2019-03-09
Add ramrodata, scratchpad linker files, and correct timebase value

bsousi5:
Dhrystone build also require LINK_TARGET=ramrodata
With this change, dhrystone on E31-Arty is now:
Microseconds for one run through Dhrystone:    6.0 
Dhrystones per Second:                      166666.7 
Which is 1.459 DMIPS


dbarbi1:
@bsousi5 one thing to note is that for the E21 and E24 there are actually 2 banks of TIMs. This will be important as in order to get the proper benchmarks, code will need to be located in one bank, and data in the other.

PR #207, opened 2019-03-14
Proposal: TARGET_TAGS is used to configure the build and to filter list-targets

nategraff-sifive:
This is my proposal to use TARGET_TAGS, implemented.

First, the Makefile now queries TARGET_TAGS to decide whether to use elf2hex, segger, etc.

Second, `make list-targets` now accepts the `TARGET_REQUIRE_TAGS` argument as a way to filter the targets by a list of values that might be in TARGET_TAGS. For example:

```
$ make list-targets TARGET_REQUIRE_TAGS="board"
bsp-list: sifive-hifive1 sifive-hifive1-revb

$ make list-targets TARGET_REQUIRE_TAGS="board jlink"
bsp-list: sifive-hifive1-revb

$ make list-targets TARGET_REQUIRE_TAGS="fpga jlink"
bsp-list:

$ make list-targets TARGET_REQUIRE_TAGS="fpga"
bsp-list: coreip-e20-arty coreip-e21-arty coreip-e24-arty coreip-e31-arty coreip-e34-arty coreip-e76-arty coreip-s51-arty coreip-s54-arty coreip-s76-arty freedom-e310-arty

$ make list-targets TARGET_REQUIRE_TAGS="rtl"
bsp-list: coreip-e20-rtl coreip-e21-rtl coreip-e24-rtl coreip-e31-rtl coreip-e34-rtl coreip-e76-rtl coreip-s51-rtl coreip-s54-rtl coreip-s76-rtl
```


nategraff-sifive:
I like the "list available target tags" option, I'll add that.

The program tags idea is interesting, but I'll hold off on that, at least for this PR.

bsousi5:
Will this handle targets in private repo or location?

nategraff-sifive:
@bsousi5 It searches for BSPs under TARGET_ROOT, so yes.

sifivekevin:
At some point we may want to consider adding some kind of "version" to the
freedom-e-sdk so higher level tools (like FS) can attempt to know about and
cope with structural changes to the SDK over time.  Otherwise it is likely
that FS will only work with versions of the SDK that were released at the
same time (and may not work with older versions that a user might point FS
to)

bsousi5:
Can it simply be "make target-tags”, instead of “list-used-target-tags” for listing available tags

nategraff-sifive:
I made it `make list-target-tags` because the other options (list-targets and list-programs) also start with "list".

bsousi5:
OK

PR #210, opened 2019-03-19
Added .map + .lst + .rtl extra output files

cgsfv:
In Freedom Studio 2018.12, there are more output files than just the .elf file:
- .map linker map output file
- .lst extended listing output file
- .hex flash binary output file

The .hex file is always generated in FS 2018.12, also for none JLink targets, for consistency.
FS 2018.12 is based on GNUMCU, so in order to give the user what he/she is used to with the new freedom-e-sdk, I suggest we add these new 3 files.

The RTL hex file output is renamed to have the file extension of .rtl, as the .hex file now always refers to the ihex format output file.


cgsfv:
Reverted to naming verilog hex files with .hex file extension.
If bsp has rtl tag, then the .hex file is in verilog hex format using elf2hex script.
If bsp does not have rtl tag, then the .hex file is in intel hex format using objcopy program.

PR #211, opened 2019-03-21
Added design.reglist files to all bsp's based on screenlog output

cgsfv:
I have converted the content in the screenlog files to Freedom Studio reglist format (design.reglist), which is simply a list of register names - one register on each line in the file.

Then I added the general purpose registers and floating point registers (where they apply) to get them included as well. For the floating point core's I also added the 3 fflags, frm and fcsr register, as they were tagged as "not supported" in the screenlog files...

I removed some registers that were not supported by GDB: ustatus, uie, utvec, uscratch, uepc, ucause, utval, uip, sstatus, sedeleg, sideleg, sie, stvec, scounteren, sepc, scause, sip, satp.

For the none-coreip bsp's I took the default list of registers from Freedom Studio.


bsousi5:
@cgsfv By using "the 3 fflags, frm and fcsr register, as they were tagged as "not supported"", if software try to access them, it will result in exception. So I am not sure it is a good idea to including them.

On organizing note, does FS need these registers in separate files?
It is possible to include this from a common directory. Like,
bsp/common/base.reglist
bsp/common/floatingpoint.reglist
bsp/common/e3-base.reglist
bsp/coreip-e31-rtl/design.reglist (include base.reglist, e3-base.reglist)

cgsfv:
Hmm - strange that you on a core with floating point (F or D) cannot access these CSR's without it resulting in an exception. How do you then access these CSR's?

It is possible to have the registers in separate files, but for now I would like to keep it simple and in separate files - 1 per bsp.

aswaterman:
mstatus.FS needs to be nonzero in order to access the floating-point state
(including the fcsr).

bsousi5:
I just pushed a change to metal and freedom-e-sdk to enable this bit.
I forgot about this bit prior to support e34 :-(

PR #212, opened 2019-03-22
Pickup floating point support from freedom-metal

bsousi5:
While at it also add PLIC fixes and only include itim region in .text section if there is no itim.


bsousi5:
Passes RTL regression run.

PR #213, opened 2019-03-22
Add a script which adds missing information to design.dts

nategraff-sifive:
Adds `scripts/fixup-dts` which looks in `design.dts` for and adds if missing:
- A PMP node
- Compat strings for global-external interrupts and local-external-interrupts
- A `sifive,testram0` node

The script is run before BSP generation if the `--no-fixup` flag is not passed to `bsp/update-targets.sh`

### DIFF
--- a/bsp/update-targets.sh
+++ b/bsp/update-targets.sh
@@ -13,10 +13,13 @@ EOF
 unset DTSFILE
 unset CUSTOM_PATH
 unset CUSTOM_NAME
+unset NO_FIXUP
+NO_FIXUP=0
 while [ "$1" != "" ]
 do
     case "$1" in
     --help)               help "$0";                                 exit 0;;
+    --no-fixup)           NO_FIXUP=1;                                shift 1;;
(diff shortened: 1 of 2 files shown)


nategraff-sifive:
@bsousi5 I created a E34-derived configuration with SCD, moving the peripheral port to 0x6000_0000 and changing it to TileLink. I've tried a few tests so far and these are the results:

Passing tests:
- hello
- return-fail (fails, therefore passes)

Failing tests (loop forever in the exception handler):
- software-interrupt
- timer-interrupt
- example-pmp

The good news though is that this script successfully generates the nodes and compat strings needed for E SDK. `metal.h` picks up the PMP and interrupt controllers, and the fact that the tests actually execute the code means that I have one data point that the heuristics for testram location are correct.

nategraff-sifive:
Rebasing on latest master fixes the problems I was seeing, so I think it's safe to say the fixup script works and the issues were caused by Metal bugs.

PR #217, opened 2019-03-26
Update the README for targets and commands

nategraff-sifive:


README.md
```diff
@@ -117,6 +125,15 @@ export RISCV_OPENOCD_PATH=/my/desired/location/openocd
...
 
+##### Install Segger J-Link Software
+
+Some targets supported by Freedom E SDK (like the SiFive HiFive1 Rev B) use
+Segger J-Link OB for programming and debugging. If you intend to use these
+targets, install the Segger J-Link Software and Documentation Pack for your
+machine:
+
+[Segger J-Link Software Downloads](https://www.segger.com/downloads/jlink#J-LinkSoftwareAndDocumentationPack)
+
```
sifivekevin: It may be worth mentioning that the Segger JLink software is included with the Freedom Studio installation, but I don't feel strongly one way or the other.

nategraff-sifive: I think these instructions are mainly for command-line users who cloned the repo manually. FS users don't need to download the toolchain or openocd either.

PR #218, opened 2019-03-26
Making dhrystone public

bsousi5:


bsousi5:
Just ran on RTL and getting the following for E31-AHB
1.60427430

dconn-sifive:
Do any other targets require RISCV_CMODEL=medany (from RISCV_CMODEL=medlow) or just E31?

bsousi5:
Typically E31 and above require this.

aswaterman:
I thought it would only be needed for RV64 cores (but I might be missing something about the linking setup here)

bsousi5:
Have not get a chance to compare the generated asm codes, but was noticing dhry numbers improved lots if using medany vs medlow 

- Bunnaroath

dconn-sifive:
Quick Coremark test on E31-arty
medany = 2.65 Coremarks / MHz
medlow = 2.89 Coremarks / MHz

This medlow/medany change would affect any example including Coremark when made public.  I see 1.5912 in RTL sim with medlow vs 1.60427430 medany (from what you mention above).  I will look at Coremark on RV64 Arty targets to see if meday helps.

palmer-dabbelt:
`-mcmodel=medany` should only matter for 64-bit cores, and then should only matter for those that have memory above physical address 2GiB (which is usually a DTIM).  It should be possible to get very similar Dhrystone and Coremark performance between medany and medlow.

PR #219, opened 2019-03-26
Rename example-coreip-welcome to sifive-welcome

nategraff-sifive:


dconn-sifive:
Looks solid, or as many would say, LGTM!

PR #224, opened 2019-03-28
Update Metal to v201903

nategraff-sifive:
- Enable RX on the sifive,uart0 driver
- Enable FP support in crt0.S
- Mark the FE310 PLL config table const

Including @sifivekevin @cgsfv so you can test the Metal fixes in Freedom Studio before merge.


nategraff-sifive:
Regressions are clean and I built a couple projects in Freedom Studio with no issues.

PR #225, opened 2019-03-29
Keep small read-only data in flash

mmicko:
In this case only metal.default.lds is affected (for others even if it is changed it will not have effect).
By this all small read-only data blocks are only store in flash leaving more memory to user.

### DIFF
--- a/bsp/coreip-e20-arty/metal.default.lds
+++ b/bsp/coreip-e20-arty/metal.default.lds
@@ -57,6 +57,12 @@ SECTIONS
 		*(.rdata)
 		*(.rodata .rodata.*)
 		*(.gnu.linkonce.r.*)
+		. = ALIGN(8);
+		*(.srodata.cst16)
+		*(.srodata.cst8)
+		*(.srodata.cst4)
+		*(.srodata.cst2)
+		*(.srodata .srodata.*)
 	} >flash AT>flash :flash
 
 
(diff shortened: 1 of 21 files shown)


ilg-ul:
I did not study your build procedures in detail, so ignore this if I got it wrong, but did you consider the possibility of splitting the linker scripts into two parts, one specific to the device and one common to all devices? In this case you would reduce redundancy and do not have to edit so many files for the same small edit.

nategraff-sifive:
@mmicko @ilg-ul The linker scripts are auto-generated by [sifive/freedom-devicetree-tools](https://github.com/sifive/freedom-devicetree-tools) based on each BSP's `design.dts`.

These changes should really be codified in the linker script generator. The linker script generator is a bit of a mess currently, though. I have a complete rewrite planned in the next month or two now that we've scoped out what we need out of it.

mmicko:
@nategraff-sifive Thanks, I understand. If you find this change to be temporary useful (till rewrite of tool is done) you can apply this, or I could apply change to current version of tool to fix this issue ?

nategraff-sifive:
If you'd like to submit a PR to the `freedom-ldscript-generator` tool we'd appreciate that. We're trying not to manually edit the generated files.

Also an FYI, we're tagging a release of Freedom E SDK today, and by popular demand I think the plan is to merge new changes to Freedom E SDK to a development branch, so expect this change to end up there once it filters up the submodules.

Closing the PR (but feel free to respond if you have comments/questions) :)

mmicko:
Created pull request for ldscript-generator https://github.com/sifive/freedom-devicetree-tools/pull/61

PR #241, opened 2019-05-03
Add user mode example programs

nategraff-sifive:
- example-user-mode
- example-user-syscall


nategraff-sifive:
Are you looking for something for humans to read or for a regression system to use for pass/fail criteria?

bsousi5:
Humman. Regression can use the return value

PR #250, opened 2019-05-13
Remove indirection in standalone projects

sifivekevin:
Copy bsp/metal.h to freedom-metal/metal/machine.h, removing the indirection that Eclipse/CDT cannot process.  I'm not advocating that this is the best solution here, but it works really well in Freedom Studio.  If there is a better way to do this feel free to "deny" PR and delete this branch.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -193,6 +193,8 @@ standalone: \
 	cp debug.mk $</debug.mk
 	cp release.mk $</release.mk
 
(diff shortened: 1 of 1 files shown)


nategraff-sifive:
@cgsfv Yeah, this should be done in the freedom-metal makefile. The generation can also be redirected to the file we want.

bsousi5:
Closing this PR as it has been addressed from freedom-metal

PR #255, opened 2019-05-22
Setup default options for dhrystone release.

reclusejack:
1. Add three compile options
a. Ground rules (Default)
b. With inline and without lto
c. Best Score

2. set default iteration number for different targets. 
Don't need to modify dhrystone Makefile for fpga tests.


scripts/standalone.mk
```diff
@@ -37,7 +37,16 @@ ifeq ($(RISCV_CMODEL),)
...
-ifeq ($(LINK_TARGET),)
+ifeq ($(PROGRAM),dhrystone)
+LINK_TARGET = ramrodata
+# Ground rules
+RISCV_XCFLAGS += -mexplicit-relocs -fno-inline
+# With inline and without lto
+#RISCV_XCFLAGS += -mexplicit-relocs -finline
+# Best Score
+#RISCV_XCFLAGS += -finline -flto -fwhole-program
+RISCV_XCFLAGS += -DDHRY_ITERS=$(TARGET_DHRY_ITERS)
```
nategraff-sifive: Why are some of these flags commented out?

Also, this should be split into two separate conditionals. The LINK_TARGET should remain here, and the RISCV_XCFLAGS should go below in the Software Flags section.

scripts/standalone.mk
```diff
@@ -37,7 +37,16 @@ ifeq ($(RISCV_CMODEL),)
...
+else
```
nategraff-sifive: Should be 
```
else ifeq ($(LINK_TARGET),)
```
to preserve the original behavior, otherwise users won't be able to select a different link target.

scripts/standalone.mk
```diff
@@ -37,7 +37,16 @@ ifeq ($(RISCV_CMODEL),)
 RISCV_CMODEL = medany
 endif
 
-ifeq ($(LINK_TARGET),)
+ifeq ($(PROGRAM),dhrystone)
+LINK_TARGET = ramrodata
```
nategraff-sifive: Do we want to force `LINK_TARGET = ramrodata`? Obviously we recommend it, but I think this conditional should be
```
ifeq ($(PROGRAM),dhrystone)
ifeq ($(LINK_TARGET),)
```
so that users can override `LINK_TARGET` if they decide to.

scripts/standalone.mk
bsousi5: The terms to use for the three options should be
Legal, Fast and Max Performance
But we should provide a make flag just like LINK_TARGET for user to select which one for their build.
The default should be Legal.

reclusejack: @nategraff-sifive We want to provide the scores as ARM did.
https://developer.arm.com/ip-products/processors/cortex-m/cortex-m0

reclusejack: That's why I put 3 options on it, I will add make flag for this~

reclusejack: I've added a make flag "DHRY_OPTION"  for this and move these to Software Flags sections.

PR #270, opened 2019-06-06
Adding QEMU BSP's for E31 and S51 targets

cgsfv:


cgsfv:
I copied these BSP's from the HiFive1 BSP and manually changed them afterwards. The qemu.cfg file does the same for QEMU that openocd.cfg does for OpenOCD, just way simpler.

sifivekevin:
Tested this in Freedom Studio and it works for me.

bsp/qemu-sifive-e31/design.dts
```diff
@@ -0,0 +1,202 @@
...
+                        reg-names = "config";
+                };
+                hfclk: clock@4 {
+                        compatible = "sifive,fe310-g000,pll";
+                        clocks = <&hfxoscout &hfroscout>;
+                        clock-names = "pllref", "pllsel0";
+                        reg = <&prci 0x8 &prci 0xc>;
+                        reg-names = "config", "divider";
+			clock-frequency = <16000000>;
+                };
```
nategraff-sifive: Should QEMU still use all of these clock sources? Are you pulling in the SiFive PRCI patch that allows it to properly emulate the clocks?

cgsfv: To my understanding the PRCI patch is already present in the 3.1 QEMU code base in https://github.com/sifive/riscv-qemu/tree/riscv-qemu-3.1

nategraff-sifive: Ah yeah, that patch is in our downstream.

Can you add a comment to the README and mention which branch of what repository you're using?

PR #277, opened 2019-06-07
Update targets.sh sh to bash

maloneymr:
One liner to fix the shabang of a file.

The `update-targets.sh` file is using `bash`-specific syntax in an `if` statement around line 42 that was throwing errors when you run it via `sh`.

### DIFF
--- a/bsp/update-targets.sh
+++ b/bsp/update-targets.sh
@@ -1,4 +1,4 @@
-#!/usr/bin/env sh
+#!/usr/bin/env bash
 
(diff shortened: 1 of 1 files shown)


maloneymr:
I don't know what branch I should base this off of. I'm working off of a pinned commit for `product-coreip-sifive`. Feel free to change the merge base if you know what the proper base should be.

nategraff-sifive:
Already merged, closing, but thanks!

PR #285, opened 2019-06-17
Reorder CFLAGS for benchmarks

nategraff-sifive:
Add benchmark flags after loading the flags for the configuration so that the benchmarks can override the configuration's flags.

### DIFF
--- a/scripts/standalone.mk
+++ b/scripts/standalone.mk
@@ -50,10 +50,8 @@ endif
 endif
 
 ifeq ($(PROGRAM),coremark)
-ifeq ($(CONFIGURATION),release)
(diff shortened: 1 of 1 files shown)


reclusejack:
For E2 series and CONFIGURATION=debug, the itim section size is not big enough when we use metal.ramrodata.lds. We will get compiling error ~

nategraff-sifive:
@reclusejack I think we have a catch-22 here where something is going to be broken no matter what behavior we make default.

@dbarbi1, do you have any input on whether it's okay to have the benchmark builds require specifying `CONFIGURATION=release`?

PR #288, opened 2019-06-20
Fix benchmark standalone builds

sifivekevin:
Hi Nate,  I fixed the standalone generator to work for coremark (and presumably dhrystone too).  Tested both coremark and (non-benchmark) sifive-welcome with these changes.


sifivekevin:
Hey Nate,  you need to do the merge. :)

PR #293, opened 2019-06-21
Update BSPs

nategraff-sifive:
Update using the latest freedom-devicetree-tools.


sifivekevin:
LGTM,  These will find there way into v201905-rc2 (and final)?

nategraff-sifive:
Yep, the goal is to put them in the next RC

bsp/sifive-hifive-unleashed/openocd.cfg
```diff
@@ -1,24 +1,30 @@
+#write_config_file
+# JTAG adapter setup
 adapter_khz     10000
 
 interface ftdi
-ftdi_device_desc "Dual RS232-HS"
-ftdi_vid_pid 0x0403 0x6010
+ftdi_device_desc "Olimex OpenOCD JTAG ARM-USB-TINY-H"
```
cgsfv: The HiFive Unleashed does not use Olimex, it uses FTDI, right?

bsp/sifive-hifive1/openocd.cfg
```diff
@@ -1,34 +1,30 @@
+#write_config_file
+# JTAG adapter setup
 adapter_khz     10000
 
 interface ftdi
-ftdi_device_desc "Dual RS232-HS"
-ftdi_vid_pid 0x0403 0x6010
+ftdi_device_desc "Olimex OpenOCD JTAG ARM-USB-TINY-H"
```
cgsfv: Same here, shouldn't it also be FTDI (Dual RS232-HS)?

cgsfv:
But besides that it looks good

nategraff-sifive:
Thanks for the catch, Carsten, we're generating openocd.cfg now so I need to update the script to fix that.

PR #298, opened 2019-07-02
Hifive1 revb readme grammar

sifivekevin:
Not sure why this PR is pickup up two commits you made...


nategraff-sifive:
Mmm, it picked up those commits because they got merged to master but not backported to v201905-branch.

I also think they're reasonable to backport, so we can probably just hit the merge button.

@bsousi5 what do you think?

bsousi5:
> Mmm, it picked up those commits because they got merged to master but not backported to v201905-branch.
> 
> I also think they're reasonable to backport, so we can probably just hit the merge button.
> 
> @bsousi5 what do you think?

My concern is the ASFLAGS, since is effect the build and outputs. As all sanity run passes, I am good with it.

nategraff-sifive:
The ASFLAGS change should only be positive: it passes `-march` and friends to the assembler.

PR #300, opened 2019-07-04
Disable Segger's ISA simulator

palmer-dabbelt:
It appears that Segger will simulate some instructions, presumably to
speed up stepping past a breakpoint.  I just spent a day or so dealing
with a bug in that simulator, so let's just disable it for now.


palmer-dabbelt:
There should probably be a check box in FS, but for now we can stick it in the "extra GDB commands" window.

PR #309, opened 2019-07-10
Replace e310 arty bsp files with the ones in the E31 release package

virtualboots:
The Arty E31 BSP files are wrong (possibly outdated). I update it with the one in the E31 prebuilt tar ball. I've tested it on Arty.


virtualboots:
I didn't know they are different thing. I'll close this PR.

PR #310, opened 2019-07-11
update the coremark options for gcc 8.3.0.

reclusejack:
@issuehsu tried different options to improve coremark score with gcc 8.3.0.
With this patch, the improvement is about 3~4% for e2/s2/e3/s5 series.


virtualboots:
I prefer keeping the original style because it is clearer to read and maintain.

For example, if we want to modify or update options, the proposed style is error prone because we need to pay attention to how the final option set is composed by different lines.

palmer-dabbelt:
From https://github.com/sifive/benchmark-coremark/commit/287f632b3e1fa3d9ea87e03a30d33ed6c1184485 we have `-O2 -fno-common -funroll-loops -finline-functions --param max-inline-insns-auto=20 -falign-functions=4 -falign-jumps=4 -falign-loops=4 --param inline-min-speedup=10`

nategraff-sifive:
Go ahead and submit another PR to backport this to v201905-branch.

PR #316, opened 2019-07-22
Fix HiFive1 Rev B DTS

nategraff-sifive:
- Adds missing devices (secondary and tertiary SPI, PWM, UART)
- Makes PLIC interrupt map to match FE310-G002 Manual


nategraff-sifive:
Also adds external low frequency psdlfaltclk and missing fields for lfrosc.

PR #318, opened 2019-07-29
Launch QEMU simulation with `make simulate`

nategraff-sifive:
Launch a QEMU simulation for the qemu-sifive-e31 and qemu-sifive-s51 targets.


nategraff-sifive:
@cgsfv I updated it to use qemu.cfg, thanks for the suggestion!

PR #342, opened 2019-08-15
add api-version target

sifivekevin:
Adds an "api-version" target that Freedom Studio can use to determine if a given instance of freedom-e-sdk is compatible with the running version of Freedom Studio.


Makefile
```diff
@@ -89,6 +89,13 @@ endif
 
 include scripts/standalone.mk
 
+#############################################################
+# Prints version message
+#############################################################
+.PHONY: version
+version:
+	@echo 2019.08
```
nategraff-sifive: I worry about hand-writing version numbers all over the place because we'll invariably forget one. Is there a way this can be automatically detected?

Makefile
```diff
@@ -151,12 +158,31 @@ clean:
...
+# incremented if the list-targets, list-target-tags.
+# list-programs, list-options, or standalone targets change
+# in any way that affectz Freedom Studio requirements.
+# A given version of Freedom Stddio may not support 
+# 'future' versions of freedom-e-sdk.  Backward support
+# should be supported.
+#############################################################
+.PHONY: api-version
+api-version:
+	@echo 1.0
```
nategraff-sifive: What API is this versioning? Can we make it more specific?

sifivekevin: This, again, would require careful manual management.  I'm just worried that, at some point, we might make changes to the recipes (or add new recipes ) that Freedom Studio relies on and unless I have some way of knowing this, Freedom Studio could be rendered quite broken if a user points to a freedom-e-sdk instance that is incompatible, or older, or newer.  FS wont be able to accurately detect or report this to the user.  Things will just break.  Having some kind of 'api version' will at least let me maintain backward compatibility, and detect when the user points to a sdk that is too new for the given version of FS.  I'm completely open to alternate solutions.  This PR is primarily just get the conversation started. :)

nategraff-sifive: My issue is primarily with the name `api-version`. Can we call it `freedomstudio-api-version`?

sifivekevin: Absolutely.  That's a much better name.

PR #346, opened 2019-08-17
DTS Fixup: Use sifive,trace as stdout-path

bradseevers:
Adding support for using trace encoder ITC 0 trace messages as an optional replacement for the UART. Alternate putc() writes character stream to the ITC 0 stimulus register which creates AAWM trace messsages with the putc() data in them. The trace decoder (dqr) can then later reconstruct the prints when it processes the trace data.


nategraff-sifive:
Thanks Brad!

This needs to go in after the PRs to Metal and DeviceTree tools, and then there's some rebasing that needs to happen (like removing the commit marked DNM).

No worries, I'll handle the mechanics of it.

bradseevers:
Thanks Nate!

-brad

PR #359, opened 2019-09-03
Pull in latest changes from master and forward-port recent changes from v201908-branch

albertchen-sifive:
bump freedom-devicetree-tools for https://github.com/sifive/freedom-devicetree-tools/pull/154


nategraff-sifive:
Went ahead and bumped both devicetree tools and metal to make sure they stayed in sync. Also made sure the BSPs were regenerated.

PR #360, opened 2019-09-04
Add example-local-vector-interrupts

bsousi5:


nategraff-sifive:
You forgot to bump the pointer in wit-manifest.json.

PR #375, opened 2019-09-20
Support for using elf2hex from sdk-utilities package

cgsfv:
When using the python based elf2hex, FreedomStudio/Eclipse detects some of the output messages as errors, which is not pleasant for users.

The sdk-utilities package is already bundled with FreedomStudio, and when FS builds a standalone project it sets the FREEDOM_SDK_TOOLS_PATH env var to point to the bundled sdk-utilities package.

With this code addition, it makes FS build use the bundled precompiled C-based elf2hex, thereby also making the build faster - which is also possible from command line builds.

### DIFF
--- a/scripts/standalone.mk
+++ b/scripts/standalone.mk
@@ -211,10 +211,16 @@ $(PROGRAM_ELF): \
 # Use elf2hex if we're creating a hex file for RTL simulation
 ifneq ($(filter rtl,$(TARGET_TAGS)),)
 .PHONY: software
+ifeq ($(FREEDOM_SDK_TOOLS_PATH),)
 $(PROGRAM_HEX): \
 		scripts/elf2hex/install/bin/$(CROSS_COMPILE)-elf2hex \
 		$(PROGRAM_ELF)
 	$< --output $@ --input $(PROGRAM_ELF) --bit-width $(COREIP_MEM_WIDTH)
+else
+$(PROGRAM_HEX): \
+		$(PROGRAM_ELF)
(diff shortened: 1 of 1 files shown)


scripts/standalone.mk
```diff
@@ -211,11 +211,17 @@ $(PROGRAM_ELF): \
 # Use elf2hex if we're creating a hex file for RTL simulation
 ifneq ($(filter rtl,$(TARGET_TAGS)),)
 .PHONY: software
+ifeq ($(FREEDOM_SDK_TOOLS_PATH),)
```
bsousi5: Rather than having this check, would it make sense to have FREEDOM_SDK_TOOLS_PATH default to scripts/elf2hex/install/bin/$(CROSS_COMPILE)-elf2hex, and FS overside this?

cgsfv: I don't think so, as this will keep the dependency to having the python elf2hex compiled, so I would rather have it like this.

Will you approve it to get in to 19.08?

nategraff-sifive:
I'm okay with this as long as we don't risk backwards-incompatibility between freedom-e-sdk and Freedom Studio if we decide to do this differently in the future.

scripts/standalone.mk
bsousi5: My immediate concern is we have two different means of producing hex files, without  the ability to know from which tool was used to produce the hex file. Of course,  if users use freedom-e-sdk cmd line, it will be from the python version and FS will use the C++ version. Unfortunately, when user report an issue, they most often don't tell us that. So to unblock FS having this issue, we need to have this. But we should let CX team to get this info some how, until the next release when we only use one version.

bsousi5: Since this is very late in the game, how critical is this? Will FS work if we don't have this? Can it goes in the next 19.08 v1 let say.

cgsfv: Yes, I am OK with it going into 19.08.1

PR #386, opened 2019-10-08
Add support for the Spike RISC-V ISA Simulator

nategraff-sifive:
Adds Spike as as supported build and simulation target.

Build and simulate in spike with `make TARGET=spike PROGRAM=hello simulate`

This leverages and extends the recent feature I added to regenerate BSP files at build time with the `bsp/spike/spike_options.sh` file. If you have `freedom-devicetree-tools` in your path and edit `bsp/spike/spike_options.sh` to set `SPIKE_OPTIONS="your spike arguments here"`, Freedom E SDK will regenerate a BSP for the target which Spike emulates when passed those options. For example, with no arguments Spike emulates a core with ISA `rv64imafdc`. If you prefer to simulate `rv32imac`, you can run

`echo "export SPIKE_OPTIONS=\"--isa=rv32imac\"" > bsp/spike/spike_options.sh`
`make TARGET=spike PROGRAM=hello simulate`

and it will do the right thing.


Makefile
```diff
@@ -358,10 +358,15 @@ simulate: $(PROGRAM_ELF)
...
 	scripts/simulate --elf $(PROGRAM_ELF) --qemu $(QEMU_RISCV64) --qemu-config bsp/$(TARGET)/qemu.cfg
-endif
+endif # findstring rv32
 else # findstring qemu
+ifeq ($(findstring spike,$(TARGET)),spike)
+simulate: $(PROGRAM_ELF)
+	scripts/simulate --elf $(PROGRAM_ELF) --spike $(shell which spike)
+else
 simulate:
 	@echo "QEMU can't simulate target $(TARGET)!"
```
bsousi5: Should this be Unknown simulator/emulator for target ...

bsp/spike/metal-inline.h
```diff
@@ -0,0 +1,139 @@
+/* Copyright 2019 SiFive, Inc */
+/* SPDX-License-Identifier: Apache-2.0 */
+/* ----------------------------------- */
+/* ----------------------------------- */
+
+#ifndef ASSEMBLY
+
+#ifndef METAL_INLINE_H
```
bsousi5: Just occurred to me, I think we should also emit the target name in these generate files to, since all metal files name will be the same.

scripts/simulate
```diff
@@ -12,6 +13,7 @@ do
     case "$1"
     in
     --elf)                   elf="$2";   shift 2;;
+    --spike)               spike="$2";   shift 2;;
```
bsousi5: Nick expressed that he also need to specify spike options/configs just like qemu.
Is that a following PR by you or Nick?

knightsifive: My hope is that `SPIKE_OPTIONS` environment variable will (a) allow `freedom-e-sdk` to discern the appropriate Device Tree and build itself appropriately (b) get passed to `spike` as command-line options. It looks to me like the current PR includes this functionality.

nategraff-sifive:
I'm still working on this. It's still rebuilding metal when SPIKE_OPTIONS don't change.

PR #390, opened 2019-10-11
Add i2c example

NandkumarJoshi:


NandkumarJoshi:
> The I2C example should use metal_i2c_get_device, not the internal-use handles.

Corrected example code.

nategraff-sifive:
One more nitpick on this example:

```
#if defined(__METAL_DT_I2C0_BAUDRATE0) && defined(__METAL_DT_I2C0_MODE0)
  ...
  metal_i2c_init(i2c, __METAL_DT_I2C0_BAUDRATE0, __METAL_DT_I2C0_MODE0);
```

`#define`s prefixed with underscores aren't public and shouldn't be used in the examples. I think the right thing here is to hard-code the speed and mode of the I2C bus into the example, because it already makes an assumption about the device it's talking to.

nategraff-sifive:
I also agree with Carsten's feedback on the `transfer` function. We should have separate buffers for receive and transmit.

NandkumarJoshi:
> I am not that deep into the details of adding support for a new peripheral.
> But, shouldn't the metal_i2c_transfer function have separate rxbuf and txbuf arguments instead of a single combined buf argument?
> From a user perspective, I would like to keep rx and tx separate.

Fixed.

NandkumarJoshi:
> One more nitpick on this example:
> #if defined(__METAL_DT_I2C0_BAUDRATE0) && defined(__METAL_DT_I2C0_MODE0)
>   ...
>   metal_i2c_init(i2c, __METAL_DT_I2C0_BAUDRATE0, __METAL_DT_I2C0_MODE0);
> 
> #defines prefixed with underscores aren't public and shouldn't be used in the examples. I think the right thing here is to hard-code the speed and mode of the I2C bus into the example, because it already makes an assumption about the device it's talking to.

Macros removed, dts cleaned up.

nategraff-sifive:
@NandkumarJoshi I think we should be good here, thanks so much for your work on this! After we merge the Metal and Devicetree Tools PRs go ahead and update the submodules/wit manifest and rebase.

NandkumarJoshi:
> @NandkumarJoshi I think we should be good here, thanks so much for your work on this! After we merge the Metal and Devicetree Tools PRs go ahead and update the submodules/wit manifest and rebase.

done.

PR #394, opened 2019-10-16
Only change compilation options without changing the definition flags.

Mel-Chen:
Using RISCV_XCFLAGS=<compilation options> in the command line to change the compilation options directly.

### DIFF
--- a/scripts/standalone.mk
+++ b/scripts/standalone.mk
@@ -150,7 +150,7 @@ else ifeq ($(DHRY_OPTION),best)
 # Best Score
 RISCV_XCFLAGS += -finline -flto -fwhole-program
 endif
-RISCV_XCFLAGS += -DDHRY_ITERS=$(TARGET_DHRY_ITERS)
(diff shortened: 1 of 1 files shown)


Mel-Chen:
> Is there a reason you're only `override`-ing one of these and not the other?

For example, when someone wants to compile Coremark with -Os compilation option, he/ she may want to pass the options by using RISCV_XCFLAGS=-Os in the command line.
If all of RISCV_XCFLAGS assignments are marked with the override flag, the RISCV_XCFLAGS will become "-Os -O2 -fno-common -funroll-loops -finline-functions ...". According to the latter overrides the previous options, Coremark is still compiled with original options instead of -Os.
So that is why I only mark one RISCV_XCFLAGS assignment that appends the -D options.

Mel-Chen:
@nategraff-sifive @bsousi5 ping

nategraff-sifive:
Sorry for the delay, thanks for the ping!

PR #405, opened 2019-10-29
Fix issue of USE_LIBC_SEGGER=0 or 1 has the same behavior

lexus0:
* Original behavior:
    always use segger library whatever USE_LIBC_SEGGER=0 or 1
* Behavior after fixing:
    using nano library when USE_LIBC_SEGGER=0.
    using segger library when USE_LIBC_SEGGER=1.
    using nano library when USE_LIBC_SEGGER not apply.


issuehsu:
If there is no define of USE_LIBC_SEGGER, will this patch work?

lexus0:
Thanks for reminding.
I've updated my patch for the situation that use doesn't apply USE_LIBC_SEGGER in this commit: 8dd2f1c

issuehsu:
RISCV_LDLIBS will be appended twice.

PR #407, opened 2019-10-31
Picolibc

keith-packard:
picolibc support - requires updated freedom-metal and freedom-devicetree-tools

### DIFF
--- a/scripts/libmetal.mk
+++ b/scripts/libmetal.mk
@@ -114,7 +114,7 @@ $(METAL_BUILD_DIR)/Makefile: \
 		--host=$(CROSS_COMPILE) \
 		--prefix=$(METAL_PREFIX) \
 		--libdir=$(METAL_LIB_DIR) \
-		--with-builtin-libgloss \
+		$(METAL_WITH_EXTRA) \
 		--with-machine-header=$(abspath $(METAL_HEADER)) \
 		--with-machine-inline=$(abspath $(METAL_INLINE)) \
 		--with-platform-header=$(abspath $(PLATFORM_HEADER))
--- a/scripts/standalone.mk
+++ b/scripts/standalone.mk
@@ -78,6 +78,15 @@ else
 $(error Unable to determine XLEN from $(RISCV_ARCH))
 endif
 
+MTIME_RATE_HZ_DEF=32768
+ifeq ($(findstring qemu,$(TARGET)),qemu)
+MTIME_RATE_HZ_DEF=10000000
+else
+ifeq ($(findstring unleashed,$(TARGET)),unleashed)
+MTIME_RATE_HZ_DEF=1000000
+endif
+endif
+
 #############################################################
 # Toolchain
 #############################################################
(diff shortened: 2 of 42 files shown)


scripts/libmetal.mk
```diff
@@ -108,6 +108,9 @@ $(METAL_LIB_DIR)/stamp: $(METAL_BUILD_DIR)/Makefile
 
 $(METAL_LIB_DIR)/lib%.a: $(METAL_LIB_DIR)/stamp ;@:
 
+$(METAL_LIB_DIR)/libmetal-pico.a: $(METAL_LIB_DIR)/libriscv__pico__metal.a
```
nategraff-sifive: If we name the library what we want in freedom-metal, we can avoid defining a rule to rename it.

scripts/standalone.mk
```diff
@@ -32,6 +32,18 @@ ifeq ($(RISCV_CMODEL),)
...
+SPEC=nano
+endif
+
+ifeq ($(SPEC),nano)
+LIBMETAL_EXTRA=-lmetal-gloss
+endif
+
+ifeq ($(SPEC),picolibc)
+LIBMETAL_EXTRA=-lmetal-pico
+endif
```
nategraff-sifive: Using SPEC internally works, but I'd prefer to have a more descriptive option on the frontend for users to actually choose their libc implementation like `RISCV_LIBC` or similar

scripts/libmetal.mk
keith-packard: yeah, I'm not sure why I had this funky name; I'll try to switch it.

scripts/standalone.mk
keith-packard: Or just LIBC? And add some error checking so if the value is set to something crazy, it exits instead of breaks.

keith-packard: And I added error checking so you get a message from make when you use an invalid libc value. I haven't added any documentation about the supported values yet.

keith-packard:
Ok, this branch has been updated to use Nate's BSP generation script.

scripts/standalone.mk
```diff
@@ -36,6 +36,24 @@ ifeq ($(RISCV_CMODEL),)
 RISCV_CMODEL = medany
 endif
 
+ifeq ($(RISCV_LIBC),)
```
bsousi5: I like this. Thanks for doing this. @knightsifive has been asking for this.

keith-packard:
> Given that pico lib is not available as part of the current/standard toolchain bundle, can you add a write up in README.md to where one can go about download and use it.

Hrm. Maybe I could publish a freedom-tools add-on for picolibc until freedom-tools provides a release that includes picolibc?

nategraff-sifive:
Always building libmetal-pico.a breaks when using newlib-nano, so we should only build libmetal-pico.a when we need it. Just pushed two commits to bump freedom metal to enable and use the extra configure option.

nategraff-sifive:
DNM yet, waiting for submodule merges.

PR #411, opened 2019-11-13
Dev freertos

e-puerto:
Last update due a a mistake fix submodule

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -100,6 +100,9 @@
 [submodule "software/example-i2c"]
 	path = software/example-i2c
 	url = https://github.com/sifive/example-i2c.git
+[submodule "software/mem-latency"]
+	path = software/mem-latency
+	url = https://github.com/sifive/benchmark-mem-latency.git
 [submodule "FreeRTOS-metal"]
 	path = FreeRTOS-metal
 	url = https://github.com/sifive/FreeRTOS-metal.git
--- a/.gitmodules
+++ b/.gitmodules
@@ -106,3 +106,6 @@
 [submodule "Segger_SystemView-metal"]
 	path = Segger_SystemView-metal
 	url = https://github.com/sifive/Segger_SystemView-metal.git
+[submodule "software/example-freertos-blinky"]
+	path = software/example-freertos-blinky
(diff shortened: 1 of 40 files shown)


nategraff-sifive:
@cgsfv I'm okay with the example for the purposes of this PR, we can add a more minimal one later.

As far as the target-specific symbol, we can refactor that in the future too.

cgsfv:
@nategraff-sifive OK - I was just thinking that the target specific stuff was the responsibility of the freedom-devicetree-tool?
And isn't the MTIME_CTRL_ADDR=0x2000000 already available in some metal.h file?

e-puerto:
HI Carsten,
Here my comment regarding your remark:

> 1. Why is it so big compared to all the other examples - it's meant to be
very simple, making it easy for the user to manage and extend it when they
use it as a template for their own project
I did not have a clear firm answer for that. From my point of view it is
not so complex. Perhaps, I can little be simplified. but please do not
forget that it is not like the freedom-metal lib, I mean it should not be
included for each example (by the way, the makefile generated by the
standalone application is huge and complex).

> 2. The bsp specific details should be in the bsp generated .h files and
not in the Makefile
I agree with you but currently i am not able thanks to the settings.mk
and./or design.dts to know the clock source of mtime counter, knowing this
value is very important in order to have a tick every 1ms. regarding the
mtime_ctrl_addr same issue because it could be part of the CLINT or PLIC
(not available on the setting.mk file).

*Emmanuel Puerto*
Application Engineer, Platform Security
*SiFive France*
E: emmanuel.puerto@sifive.com
P: +33 681-359-346 <+33%20681-359-346>
W: www.sifive.com

cgsfv:
@e-puerto I know it's work in progress, but I just wanted to give you my inputs. Did you see my try-out result in Freedom Studio on Slack?

bsousi5:
A good starting to get into freedom-e-sdk. We can polish up the Makefile being FDT to generate much of the info needed in subsequence commit, in particular, freertos need to be able to run in RTL environment across different design.

Did you get a chance to validate on a E31 standard core configuration?

PR #420, opened 2019-12-05
Switch to Python-Based Tooling

nategraff-sifive:
Adds devicetree overlay generation using [devicetree-overlay-generator](https://github.com/sifive/devicetree-overlay-generator).

The Devicetree for the target design shall now reside in `bsp/TARGET/core.dts` and will be `/include/`ed into the top-level Devicetree in `bsp/TARGET/design.dts`. This top level can be used to specify the specific configuration of the target, including:
- Reset vector
- stdout-path
- Additional hardware (for example, test rams in the RTL simulation testbench)

This PR introduce a dependency on Python 3 to Freedom E SDK, as well as the packages specified in `requirements.txt`.

Marking **DNM** Until we're ready to commit to all of the overlay-generation PRs in one swell foop. Additionally, the submodules bumped in this PR should be rolled up and updated here prior to merge.


sifivekevin:
Thanks for the heads up on this!  FS will need a little work to adapt to these changes.

nategraff-sifive:
Questions to answer for this PR before merge:
1. How can we cache the Python package wheels in Freedom E SDK so that end-users don't require an internet connection to support these changes?
2. Should the design.dts Devicetree overlay be regenerated automatically at all during `make`? Which is more confusing: the overlay doesn't update automatically to reflect changes in core.dts or the overlay gets generated automatically, undoing user changes to the overlay?
3. What still needs to change for Freedom Studio support?

nategraff-sifive:
Just pushed some commits to answer Question 1 I just posted. All Python package dependencies for all platforms now live in `pip-cache`.

nategraff-sifive:
This branch now includes:
- sifive/ldscript-generator
- sifive/cmsis-svd-generator, courtesy of Carsten

The linker script generation should be a lot more consistent and controllable, especially now that layout heuristics are isolated in the overlay generator, and the user now has complete control of the layout simply by editing the Devicetree.

I also improved the Python package caching mechanism to use stock Pip commands.

Remaining freedom-devicetree-tools programs:
- freedom-metal_header-generator
- freedom-bare_header-generator
- freedom-openocdcfg-generator
- freedom-makeattributes-generator

The latter two are easy pickings for pydevicetree-based Python script rewrites, so I think those are up next. In the meantime, I welcome review and feedback on what currently exists.

cgsfv:
It actually builds fine under Freedom Studio now - that is if a C compiler is installed, due to the fact that the following shared objects are build as part of the pip install step:

./markupsafe/_speedups.cpython-37m-darwin.so
./wrapt/_wrappers.cpython-37m-darwin.so
./typed_ast/_ast27.cpython-37m-darwin.so
./typed_ast/_ast3.cpython-37m-darwin.so
./lazy_object_proxy/cext.cpython-37m-darwin.so

This kind of breaks the advantage of converting the FDTT tools from C code to Python code...
We should find a way to only need pip packages that do not need a C compiler.

nategraff-sifive:
We can fix this if we can figure out how to get Pip to download the precompiled platform-specific packages from the Python package index.

Looking at the `pip-cache` target in `scripts/virtualenv.mk`.

nategraff-sifive:
I was able to remove wrapt and lazy-object-proxy by removing pylint from freedom-e-sdk. The linter isn't necessary for the SDK, I included it because the scripts themselves use it in CI.

MarkupSafe and typed-ast remain necessary. However, downloading all the packages takes a long time and makes us bad neighbors for the Python Package Index, so I'm going to store all the cached packages in our own repository and submodule it into the SDK.

sifivekevin:
does the pip-cache need to be copied to the standalone project as part of `make standalone`?  I get build errors: Url 'pip-cache' is ignored. It is either a non-existing path or lacks a specific scheme

nategraff-sifive:
You're right, I missed that part. Thanks!

sifivekevin:
The good news: Latest commit is working on Windows
And now the bad news. On my workstation:

1. Creating and building a new standalone project in FreedomStudio take 120ish seconds with the new stuff, vs 40ish seconds with the 19.08 fesdk.  Most of that time is python environment stuff.
2. Subsequent builds (after a clean) are much faster at around 40 seconds.  So not so bad...

nategraff-sifive:
I can make the BSP file generation time faster [by increasing the parser memoization cache](https://github.com/sifive/pydevicetree/pull/42).

Not sure what I can do about the virtualenv creation time though. That might be something we're stuck with?

sifivekevin:
We should definitely pick off any low hanging fruit.  

And...I always edit out the long build times in my video's.  :)

nategraff-sifive:
Added esdk-settings-generator and openocdcfg-generator, which replace freedom-makeattrib-generator and freedom-openocdcfg-generator respectively.

scripts/virtualenv.mk
```diff
@@ -0,0 +1,41 @@
+
```
mmjconolly: Would it be applicable to use wake's python virtual env support from api-languages-sifive/python.wake

nategraff-sifive: Make is first-class in freedom-e-sdk because it's what our customers have and want.

Does api-languages-sifive handle the "cache all prebuilt wheels for common platforms" flow we need here to make sure that customers don't have to hit PyPI or need a C compiler to create the virtualenv?

PR #422, opened 2019-12-10
Add FreeRTOS Minimal and Blinky-SystemView Examples

e-puerto:
Add 2 Example (minimal & blinky+systemview from segger).

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -115,3 +115,6 @@
 [submodule "software/example-freertos-minimal"]
 	path = software/example-freertos-minimal
 	url = https://github.com/sifive/example-freertos-minimal.git
+[submodule "software/example-freertos-blinky-systemview"]
+	path = software/example-freertos-blinky-systemview
+	url = https://github.com/sifive/example-freertos-blinky-systemview.git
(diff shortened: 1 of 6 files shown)


sifivekevin:
Emmanuel, will these examples run in Freedom Studio without Segger SystemView?

cgsfv:
I tried to create a new project in Freedom Studio using the example-freertos-minimal, and gets this result in "make all CONFIGURATION=debug" after FS have created the standalone project:
image

It doesn't seem like it builds and installs the freertos kernel. I would expect it to install the header files and the library into the highlighted directories below, just like the metal headers and libraries, or?
image

e-puerto:
Issue with example-freertos-minimal fixed for standalone target.

.gitmodules
```diff
@@ -112,3 +112,9 @@
 [submodule "software/example-freertos-blinky"]
 	path = software/example-freertos-blinky
 	url = https://github.com/sifive/example-freertos-blinky.git
+[submodule "software/example-freertos-minimal"]
```
bsousi5: In the readme, can you add a comment indicating this example is minimal starter for their FreeRTOS application

PR #425, opened 2020-01-12
Pick up Memory address can exceed 32bits fix

bsousi5:


bsousi5:
> Inside if check (address < upper_bound), 'address' is limited to 32bits, will always evaluate to true for values exceeding 32bits.

The address should not exceed 32bits for RV32. For RV64 it will be 64bits.

PR #428, opened 2020-01-30
Updating svd files and generation from review comments

cgsfv:
- All lower case names
- Leave out the displayName
- Leave out reserved registers and fields
- Use freedom-e-sdk naming standard for peripheral names
- Add empty register files for interrupt controllers (first step)


nategraff-sifive:
@cgsfv when you bump a submodule, you also need to modify the commit hash in `wit-manifest.json` so that wit workspaces which depend on freedom-e-sdk can get the same version as the git submodule.

PR #433, opened 2020-03-09
Create Docker image and use for testing

nategraff-sifive:


nategraff-sifive:
DNM until I have a Personal Access Token so that merges to `master` automatically publish freedom-e-sdk:latest.

nategraff-sifive:
On second thought, we should merge this now to improve our CI flow. We can re-enable the publish-docker workflow when we have the API token story worked out. In the meantime it should still have a "post-merge" style action run on every push to master which builds the container. When we have an API token it's simple to turn on the push-to-registry steps of the publish-docker workflow.

PR #434, opened 2020-03-09
Bump FreeRTOS-metal, so bump also FreeRTOS examples

e-puerto:
Update FreeRTOS branch on top of the master.


e-puerto:
FreeRTOS have been updated in order to be updated to FreeRTOS 10.2.1, then support the vPortEndScheduler function, put the heap of FreeRTOS inside the heap of freedom metal, put the stack for the ISR management also into the heap.

PR #437, opened 2020-03-11
Dev freertos

e-puerto:
This pull request should be merge after pull request on esdk-settings-generator #3
It will permit to pass extra parameter to example-freertos-blinky. It is needed to pass parameter from setting.mk to application (in case of RTL)


nategraff-sifive:
I just merged esdk-settings-generator; bump the pointer here and I'll approve and merge this 👍

PR #439, opened 2020-03-11
Bump FreeRTOS-metal in order to support FreeRTOS running into all mhartid

e-puerto:
Change to support running FreeRTOS in any mhartid


nategraff-sifive:
Confirmed fixes the freertos examples on u54mc.

PR #440, opened 2020-03-12
Update website URL in README

richardxia:
I noticed that this is pointing at our old developers website. Although there's a redirect from the old domain to the new one, it's best to update this to point to the new one.


richardxia:
> Should we point at SCS or just at sifive.com?

Good call, yes, I think the main website would be better, since that's actually where the content described above is linked from.

PR #445, opened 2020-03-19
Add FREEDOM_E_SDK_VENV_PATH to move virtualenv

nategraff-sifive:
The FREEDOM_E_SDK_VENV_PATH environment variable allows the path to the virtualenv to be moved at will while defaulting to the same directory an freedom-e-sdk.

This is added to facilitate running Freedom E SDK in wake, where multiple BSPs might be created at the same time. By using a different virtualenv for each BSP creation job, we can eliminate contention over the virtualenv directory.

### DIFF
--- a/bsp/update-targets.sh
+++ b/bsp/update-targets.sh
@@ -78,6 +78,10 @@ fi
 
 OVERLAY_GENERATOR=../scripts/devicetree-overlay-generator/generate_overlay.py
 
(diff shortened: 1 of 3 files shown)


nategraff-sifive:
@cgsfv @sifivekevin This might be useful to you if you want to let projects in Freedom Studio share a virtualenv.

PR #447, opened 2020-03-23
Fetch Python packages from PyPI by default, keep pip-cache option

nategraff-sifive:
This PR makes fetching Python packages from PyPI the default (again).

Users can elect to prefetch all required Python packages for all supported platforms and Python versions by running `make pip-cache`, which downloads all of the packages into `$FREEDOM_E_SDK_PIP_CACHE_PATH` (by default equal to `freedom-e-sdk/pip-cache`).

### DIFF
--- a/README.md
+++ b/README.md
@@ -240,6 +240,47 @@ git pull origin master
 git submodule update --init --recursive
 ```
 
+### Python ###
+
+Freedom E SDK includes a number of Python scripts used during the build process
+to parameterize the build of Freedom Metal to the target. The dependencies of
+these scripts are tracked in `requirements.txt`. Freedom E SDK manages its own
+virtualenv, but there are some options which allow users to configure the
+virtualenv to best suit your needs.
+
+#### Predownloading Python Dependencies ####
+
+By default, Freedom E SDK will download Python packages from the Python Package
+Index when it creates the virtualenv. If you prefer to download dependencies
+ahead-of-time, you can run
+
+```
+make pip-cache
+```
+
+to download all Python packages. This mechanism downloads all of the dependencies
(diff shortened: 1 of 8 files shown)


scripts/download-python-packages.py
```diff
@@ -0,0 +1,23 @@
...
+    destination = sys.argv[2]
+
+    platforms = ["manylinux1_x86_64", "macosx_10_13_x86_64", "win32", "win_amd64"]
+    versions = ["35", "36", "37"]
+
+    tuples = product(platforms, versions)
+
+    base_cmd = ["python3", "-m", "pip", "download", "-d", destination, "-r", requirements]
+
+    subprocess.run(base_cmd + ["--platform", "any", "--no-deps"])
```
richardxia: Beware, the default behavior of `subprocess.run()` on failure is to not handle it. You can set [`check=True`](https://docs.python.org/3.5/library/subprocess.html#subprocess.run) if you want it to throw an exception on non-zero exit. Alternatively, you can handle it manually by inspecting the [`subprocessCompletedProcess`](https://docs.python.org/3.5/library/subprocess.html#subprocess.CompletedProcess) object that `run()` returns.

sifivekevin:
I'm getting this error when attempting to create and build a stand-alone project:

`make: *** No rule to make target 'pip-cache', needed by 'standalone'.  Stop.`

Maybe standalone.mk needs updates too?

README.md
```diff
@@ -240,6 +240,47 @@ git pull origin master
 git submodule update --init --recursive
 ```
 
+### Python ###
+
+Freedom E SDK includes a number of Python scripts used during the build process
+to parameterize the build of Freedom Metal to the target. The dependencies fo
```
dbarbi1: typo, fo = of

PR #449, opened 2020-03-24
Freertos upd cfg

phmoussay:
Change on FreeRTS lds, modification of FreeRTOS auto generated config header and change to compile System View as a library


.gitmodules
```diff
@@ -139,3 +139,6 @@
 [submodule "scripts/esdk-settings-generator"]
 	path = scripts/esdk-settings-generator
 	url = https://github.com/sifive/esdk-settings-generator.git
+[submodule "software/example-freertos-pmp-blinky"]
+	path = software/example-freertos-pmp-blinky
+	url = git@github.com:sifive/example-freertos-pmp-blinky
```
nategraff-sifive: Submodule URL need to be in HTTPS format to allow unauthenticated users to clone them.

phmoussay: Fixed it.

bsousi5: Since this is an example, can you move to where the rest of the examples are. Make it easier to trace later.

bsousi5:
> Isn't there suppose to be a Makefile or standalone.mk file change for this? So that all freertos examples uses metal.freertos.lds

Nevermind. Just saw where the changes is. GitHub slow to upload

Issue #461, opened 2020-04-12
AON blocks in core devicetrees have base-address marked as `mem` instead of `control`


sctincman:
Most parent blocks appear to reference the block base-address via `regs: <{addr} {size}>` and mark them as such via `reg-names: "control"`, however the aon block in the FE310/FS510 trees instead refers to its base as `reg-names: "mem"`.

This doesn't affect `rtc` and `wdog` as they are derived from the aon block itself and simply use the base-address emitted, but the lfclk/lfrosc entry/device in freedom-devicestree-tools specifically matches on the reference's `control` reg-name. As aon here doesn't define `control` this will return 0 and cause an incorrect entry for the two register addresses used in lfclk/lfrosc, which is partially responsible for https://github.com/sifive/freedom-metal/issues/238 (get_rate_hz on lfclk returns 0).

`reg-name: "control"` here seems appropriate and used in other blocks (and mem appears to be reserved for things like DTIM/ITIM).

Affected DTS:
bsp/qemu-sifive-e31/core.dts
bsp/qemu-sifive-s51/core.dts
bsp/sifive-hifive1/core.dts
bsp/sifive-hifive1-revb/core.dts

PR #462, opened 2020-04-12
Change aon block reg-name to "control"

sctincman:
Addresses #461, verified correct register address generated in metal.h after rebuilding bsp.


sctincman:
Required updates to cmsis-svd-generator and freedom-devicetree-tools (PRs https://github.com/sifive/freedom-devicetree-tools/pull/211 and https://github.com/sifive/cmsis-svd-generator/pull/3). Updated submodules and wit-manifest to point to child PR commits

PR #466, opened 2020-04-16
Standalone venv

sifivekevin:


sifivekevin:
So the original code is:
```
# Set FREERTOS_METAL_VENV_PATH to re-use Freedom E SDK's virtualenv
export FREERTOS_METAL_VENV_PATH = $(FREEDOM_E_SDK_VENV_PATH)
```
But FREEDOM_E_SDK_VENV_PATH is either not defined, or "" when the standalone make is run.  Are you saying just do something like this:
```
# Set FREERTOS_METAL_VENV_PATH to re-use Freedom E SDK's virtualenv
export FREEDOM_E_SDK_VENV_PATH = venv
export FREERTOS_METAL_VENV_PATH = $(FREEDOM_E_SDK_VENV_PATH)
```

scripts/standalone.mk
```diff
@@ -10,9 +10,12 @@ SRC_DIR ?= $(abspath src)
 export FREERTOS_SOURCE_PATH = $(abspath FreeRTOS-metal)
 # FREERTOS_SOURCE_PATH sets the path to the SEGGER SystemView source directory
 export SYSTEMVIEW_SOURCE_PATH = $(abspath Segger_SystemView-metal)
-# Set FREERTOS_METAL_VENV_PATH to re-use Freedom E SDK's virtualenv
+# Set FREEDOM_E_SDK_VENV_PATH to use a project-local virtualenv
+export FREEDOM_E_SDK_VENV_PATH = venv
```
nategraff-sifive, line 14: Hmm. I think what's still happening here is that since `venv` isn't an absolute path, the default behavior this results in is for Freedom E SDK components to make a virtualenv in `path/to/freedom-e-sdk/venv` and FreeRTOS-metal makes a virtualenv in `path/to/freedom-e-sdk/FreeRTOS-metal/venv`.

My goal would be to make them both use `path/to/freedom-e-sdk/venv` by default. Then:

- if only FREEDOM_E_SDK_VENV_PATH is set in the environment, both use FREEDOM_E_SDK_VENV_PATH
- if only FREERTOS_METAL_VENV_PATH is set in the environment, FreeRTOS-metal uses FREERTOS_METAL_VENV_PATH
- if both are set in the environment, they both use their own virtualenvs according to their respective environment variables.

nategraff-sifive: And then in the context of standalone projects, the default should be `path/to/standalone/project/venv`

PR #467, opened 2020-04-16
Test standalone project creation and build in GitHub Actions

nategraff-sifive:
Currently tests `hello` and `example-freertos-minimal`.


nategraff-sifive:
> I have no idea how to test this but it looks great to me!

Ideally, we sit here and wait for GitHub to show us a green checkmark. :)

PR #472, opened 2020-04-24
scl-metal librarie (driver for HCA)

e-puerto:
Do not merge this before PR on freedom-devicetree-tools #212.
This PR will add necessary to add HCA usage on freedom-e-sdk.
- freedom-devicetree-tools will generated the HEADER for the HCA version 0.5.x
- scl-metal contain HCA driver
- software/example-hca-metal: example to low level access to HCA.

### DIFF
--- a/bsp/freedom-e310-arty/metal.h
+++ b/bsp/freedom-e310-arty/metal.h
@@ -861,11 +861,6 @@ struct __metal_driver_cpu *__metal_cpu_table[] = {
 
 #define __METAL_DT_INTERRUPT_CONTROLLER_C000000_HANDLE (&__metal_dt_interrupt_controller_c000000.controller)
 
-/* From local_external_interrupts_0 */
-#define __METAL_DT_SIFIVE_LOCAL_EXINTR0_HANDLE (&__metal_dt_local_external_interrupts_0.irc)
-
-#define __METAL_DT_LOCAL_EXTERNAL_INTERRUPTS_0_HANDLE (&__metal_dt_local_external_interrupts_0.irc)
-
 #define __MEE_DT_MAX_GPIOS 1
 
 __asm__ (".weak __metal_gpio_table");
--- a/.gitmodules
+++ b/.gitmodules
@@ -150,7 +150,7 @@
 	url = https://github.com/sifive/example-buserror.git
 [submodule "software/example-hca-metal"]
 	path = software/example-hca-metal
-	url = git@github.com:sifive/example-hca-metal.git
+	url = https://github.com/sifive/example-hca-metal.git
 [submodule "scl-metal"]
 	path = scl-metal
-	url = git@github.com:sifive/scl-metal.git
+	url = https://github.com/sifive/scl-metal.git
--- a/.gitmodules
+++ b/.gitmodules
@@ -148,3 +148,9 @@
 [submodule "software/example-buserror"]
 	path = software/example-buserror
 	url = https://github.com/sifive/example-buserror.git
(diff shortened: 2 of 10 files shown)


nategraff-sifive:
Why aren't we integrating the HCA driver into the Metal codebase?

bsp/freedom-e310-arty/metal-platform.h
```diff
@@ -125,4 +125,6 @@
 #define METAL_SIFIVE_UART0_IP 20UL
 #define METAL_SIFIVE_UART0_DIV 24UL
 
+#define METAL_SIFIVE_HCA
```
bsousi5, line 128: Why are these get added if the feature does not appears in the dts file?

bsp/freedom-e310-arty/metal.h
```diff
@@ -861,11 +861,6 @@ struct __metal_driver_cpu *__metal_cpu_table[] = {
 
 #define __METAL_DT_INTERRUPT_CONTROLLER_C000000_HANDLE (&__metal_dt_interrupt_controller_c000000.controller)
 
-/* From local_external_interrupts_0 */
-#define __METAL_DT_SIFIVE_LOCAL_EXINTR0_HANDLE (&__metal_dt_local_external_interrupts_0.irc)
```
bsousi5, line 865: Why was these get deleted?

nategraff-sifive: I removed them in a recent fix for freedom-devicetree-tools. They weren't being used by anything, were private API, and were generating compile warnings because multiple were getting generated which conflicted with each other.

bsousi5:
Oh OK. Just that we have not regenerate/update the BSP files in Freedom-e-sdk. Thanks

bsp/freedom-e310-arty/metal.h
bsousi5: Oh OK. Just that we have not regenerate/update the BSP files in Freedom-e-sdk. Thanks

bsp/freedom-e310-arty/metal-platform.h
e-puerto: look like an issue with the freedom-devicetree-tools generation. I should probably remove it, because this define is not used ;)

nategraff-sifive:
If we're committed to keeping the HCA driver separate from Metal, then we need to think through the entire flow including how it's built during IP delivery creation and how it's built in the customer's environment.

Obviously it's easier if the driver is completely integrated into Metal because we've already answered all of those questions for Metal. So what I'd like to know is if there's a good reason to keep it separate.

e-puerto:
HCA driver is not the Final target. the final target is le lib SCL having
the HCA "driver" permit to propose a software alternative if the HCA is not
present, but also to have multiple instances of HCA
What do you mean by "metal" do you mean put all into lib-metal?

*Emmanuel Puerto*
Application Engineer, Platform Security
*SiFive France*
E: emmanuel.puerto@sifive.com
P: +33 681-359-346 <+33%20681-359-346>
W: www.sifive.com

nategraff-sifive:
> HCA driver is not the Final target. the final target is le lib SCL having the HCA "driver" permit to propose a software alternative if the HCA is not present, but also to have multiple instances of HCA What do you mean by "metal" do you mean put all into lib-metal?

@e-puerto Don't worry, scl-metal being separate is just fine now that I've thought about it a bit. Carry on! :)

bsousi5:
Once pre-merge CI are passing, can you clean up/squash the commit logs. Thanks

PR #473, opened 2020-04-26
Update BSPs for ecc memory scrub. Bump submodules.

NandkumarJoshi:


nategraff-sifive:
So the error that happened here was because the ldscript generator didn't guarantee reproducibility. I've fixed the problems and already bumped the generators past these commits, so you can probably get away with reducing this PR to just bump freedom-metal.

bsousi5:
You also need to bump freedom-e-sdk in wit-manifest.json. I merged a fix to devicetree-overlay-generator this afternoon.

NandkumarJoshi:
> So the error that happened here was because the ldscript generator didn't guarantee reproducibility. I've fixed the problems and already bumped the generators past these commits, so you can probably get away with reducing this PR to just bump freedom-metal.

Thanks for the fix ! Updated this PR to pickup latest freedom-metal, device-tools, ld-generator. Regenerate BSPs.

debug.mk
```diff
@@ -9,6 +9,5 @@ RISCV_CXXFLAGS += -O0
 
 # Enable debug
 RISCV_ASFLAGS += -g
-RISCV_CFLAGS += -g
-RISCV_CXXFLAGS += -g
-
+RISCV_CFLAGS += -g -mno-relax
+RISCV_CXXFLAGS += -g -mno-relax
```
bsousi5, line 13: Since this change is limited to only with debug build, I will allow this for now. We would not want to do this across the board.

PR #481, opened 2020-05-04
Fix release notes generation with GitHub Action

nategraff-sifive:
DNM until submodules are merged.


.github/workflows/release.yml
```diff
@@ -40,14 +40,11 @@ jobs:
           ./scripts/tag-submodules.sh ${tag}
 
       - name: 'Create Release Notes'
+        uses: sifive/action-release-notes@v1
```
bsousi5, line 43: Here the release version will be enforce by the tag format defined above.
I assume any entity/person can uses this action, does not need to conform to our format for their release?

nategraff-sifive: Yeah, the action accepts any tag you want to provide it.

PR #486, opened 2020-05-06
Bump all example-freertos

e-puerto:
Remove metal_interrupt_enable call into all FreeRTOS_init function in all FreeRTOS example.
It will avoid random failure into RTL simulation.

### DIFF
--- a/wit-manifest.json
+++ b/wit-manifest.json
@@ -210,17 +210,17 @@
         "source": "git@github.com:sifive/Segger_SystemView-metal.git"
     },
     {
-        "commit": "66d8426f84c526a552933795fb32c3c831ea3165",
+        "commit": "71335eb86df37684a962524a833d26ffd9041599",
         "name": "example-freertos-blinky",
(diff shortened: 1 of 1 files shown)


wit-manifest.json
```diff
@@ -235,7 +235,7 @@
         "source": "git@github.com:sifive/example-clic-nested-interrupts.git"
     },
     {
-        "commit": "27a9b3b44b07598e8be6492edb27c09b611eec8f",
+        "commit": "2b5425aac870ac892ed1e08134b00a6e0184a062",
```
bsousi5, line 238: This update also picked "add check on pmp granularity" changes. I hope you already regress this on all cores

e-puerto: exactly,  the example-freertos-pmp-blinky, check granularity add the number of PMP, if it is not compliant with the program we will print a message on the uart and exit with 0.

bsousi5: ok. A nop for non-compliance case, which should gracefully exit

PR #489, opened 2020-05-11
U8 initial support

bsousi5:


scripts/standalone.mk
```diff
@@ -153,8 +153,13 @@ RISCV_CFLAGS    += -march=$(RISCV_ARCH) -mabi=$(RISCV_ABI) -mcmodel=$(RISCV_CMOD
 RISCV_CXXFLAGS  += -march=$(RISCV_ARCH) -mabi=$(RISCV_ABI) -mcmodel=$(RISCV_CMODEL)
 RISCV_ASFLAGS   += -march=$(RISCV_ARCH) -mabi=$(RISCV_ABI) -mcmodel=$(RISCV_CMODEL)
 # Prune unused functions and data
+ifeq ($(PROGRAM),dhrystone)
+RISCV_CFLAGS   += -fno-function-sections -fno-data-sections
+RISCV_CXXFLAGS += -fno-function-sections -fno-data-sections
```
nategraff-sifive, line 158: This is going to break our current/recently-added method of reducing ITIM consumption by linking only the right function sections into the ITIM.

bsousi5: > This is going to break our current/recently-added method of reducing ITIM consumption by linking only the right function sections into the ITIM.

Ah yes I forgot about that. We need to address that going forward. For now, I will limit this to on 8-series then.

keith-packard:
We can switch the current hack to be filename based, which should be fine?

scripts/standalone.mk
```diff
@@ -204,22 +214,29 @@ RISCV_XCFLAGS += -DDHRY_ITERS=$(TARGET_DHRY_ITERS)
 endif
 
 ifeq ($(PROGRAM),coremark)
+ifeq ($(RISCV_SERIES),sifive-8-series)
 ifeq ($(RISCV_SERIES),sifive-7-series)
```
nategraff-sifive, line 218: This will never be true because `RISCV_SERIES` will never be both `sifive-8-series` and `sifive-7-series`

scripts/standalone.mk
nategraff-sifive: As long as the U8 doesn't need code in ITIM, this at least won't break non 8-series targets.

jsmithsf: 8-series does not support TIMs of any sort (neither I nor D).

scripts/standalone.mk
bsousi5: Added a separate check for 8-series to use 7-series mtune for now, which can be changed in the future

PR #491, opened 2020-05-13
Picking Dhrystone strcmp.S, MC interrupt fix

bsousi5:


bsousi5:
Internal CI Run for Benchmark-Scores-Regression #5 now is good

keith-packard:
Note that picolibc *always* builds the optimized version of strcmp so that dhrystone wouldn't need a locally optimized version :-)

PR #493, opened 2020-05-13
Bump to pick up Dhrystone, MC interrupt fix

bsousi5:


bsousi5:
Will merged this without approval as it has been approved and merged into v20.05-branch already

PR #495, opened 2020-05-15
V20.05 with gcc10

issuehsu:


bsousi5:
This should go into master and not into v20.05-branch. v20.05-branch is a Koala release branch that has been qualify with 19.08 and in progress a rebuild of gcc-8.3 with limited patches and fixes per Freedom-tools v20.04.

issuehsu:
OK, I will resend the PR

PR #498, opened 2020-05-18
Regen BSPs. Fix for gcc10.

NandkumarJoshi:


NandkumarJoshi:
@issuehsu Multiple definition errors with GCC10 in metal code base are fixed. '-fcommon' option is not needed for successful compilation.

issuehsu:
Cool! I will remove -fcommon if all the multiple definition errors are fixed.

PR #499, opened 2020-05-19
Bump ldscript-generator for scrub limit to 64KB

bsousi5:


bsousi5:
Accidentally pulled against the wrong branch

PR #501, opened 2020-05-20
Refactor and Add Wake API to Freedom E SDK

nategraff-sifive:
Here is my proposal for bringing the Freedom E SDK Wake API into Freedom E SDK. In doing so, I've also made some modifications in service of improving, clarifying, and making the API more maintainable.

### FreedomESDKTarget

`FreedomESDKTarget` is now the tuple for adding targets to Freedom E SDK and building programs. A `FreedomESDKTarget` can be created with `makeFreedomESDKTargetFromCoreDTS`, which I've named to clarify that we generate the design.dts overlay ourselves and to permit the creation of new APIs which provide the Devicetree overlay in the future. The comment/"function signature" from the code is reproduced below:

```
# makeFreedomESDKTargetFromCoreDTS creates a FreedomESDKTarget given:
#  - sdk: an instance of FreedomESDK, created by makeFreedomESDK
#  - targetName: the name of the target as a String
#  - coreDTS: a single Devicetree Source file which describes the hardware
#             of the target device. This DTS will be used to generate a
#             Devicetree Overlay
#  - targetType: one of FreedomESDKTargetType, used to configure the BSP
#                file generators.
```

Consumers of the Freedom E SDK Wake API will need to do one of the following to use Freedom E SDK:
 1. Adopt `FreedomESDKTarget` as their representation of their build targets (probably least likely or appropriate).
 2. Wrap `FreedomESDKTarget` in their own build target representation.
 3. Define a mapping between their own build target representation and `FreedomESDKTarget`.

### FreedomESDKProgram

Freedom E SDK now owns the list of programs it can build and imposes its own filtering rules when queried for that list of programs. Note in programs.wake:

```
# Get the list of Freedom E SDK examples which are compatible with a given target
global def allFreedomESDKExamples sdkTarget =
  def programList =
    makeFreedomESDKProgram "hello" "example",
    ... TRUNCATED ...
    makeFreedomESDKProgram "uart-interrupt" "example",
    Nil
  filterFreedomESDKPrograms sdkTarget programList
```

The list of examples is a function of the `FreedomESDKTarget`. This reduces the risk that someone will attempt to build an example which is incompatible with the target. Consumers of this API can optionally extend or additionally filter the returned list of programs, or ignore the list entirely.

### DIFF
--- a/.github/workflows/wake.yml
+++ b/.github/workflows/wake.yml
@@ -14,7 +14,7 @@ jobs:
           path: wit-packages/freedom-e-sdk
 
       - name: 'Init Wit Workspace'
-        uses: sifive/wit/actions/wit@v0.13.0
+        uses: sifive/wit/actions/wit@9fdc590
         with:
           command: init
           arguments: |
@@ -31,4 +31,4 @@ jobs:
             --workdir="/mnt/workspace" \
             --rm \
             sifive/environment-blockci:0.4.0
-            wake -v -x testFreedomESDKWakeAPI Unit
+            wake -v --in freedom_e_sdk_test -x testFreedomESDKWakeAPI Unit
--- a/.github/workflows/wake.yml
+++ b/.github/workflows/wake.yml
@@ -1,0 +1,34 @@
+name: wake-api-test
+
+on: push
+
+jobs:
+  test:
+    runs-on: ubuntu-latest
+    env:
+      WIT_WORKSPACE: wit-workspace
+
+    steps:
+      - uses: actions/checkout@v2
+        with:
+          path: wit-packages/freedom-e-sdk
+
+      - name: 'Init Wit Workspace'
+        uses: sifive/wit/actions/wit@v0.13.0
+        with:
+          command: init
+          arguments: |
+            ${{ env.WIT_WORKSPACE }}
+            -a ./wit-packages/freedom-e-sdk
+            -a git@github.com:sifive/environment-blockci-sifive.git::0.4.0
+          force_github_https: true
+
+      - name: 'Run Wake API Test'
+        run: |
+          set -euo pipefail
+          docker run \
+            --volume="$GITHUB_WORKSPACE/$WIT_WORKSPACE:/mnt/workspace" \
+            --workdir="/mnt/workspace" \
+            --rm \
+            sifive/environment-blockci:0.4.0
+            wake -v -x testFreedomESDKWakeAPI Unit
--- a/build.wake
+++ b/build.wake
@@ -63,39 +63,51 @@ def filterWakeAndDotFiles =
   | filter (!matches `.*\.wake` _.getPathName)
 
 #
-# Plan Resources
+# Build Environment
 #
+
+export tuple FreedomESDKBuildEnv =
+  export Resources: List String
+  export BuildDir: String
 
 export def defaultFreedomESDKResources =
   "riscv-tools/2019.08.0",
   "python/python/3.7.1",
   Nil
 
+export def defaultFreedomESDKBuildDir = "build/freedom-e-sdk"
+
+export def defaultFreedomESDKBuildEnv =
+  FreedomESDKBuildEnv defaultFreedomESDKResources defaultFreedomESDKBuildDir
+
 #
 # Virtualenv Creation
 #
 # Freedom E SDK depends on a number of Python libraries. Here we enable
-# the installation of a single, workspace-wide virtualenv which can be
-# preinstalled using either the `preinstallFreedomESDKVirtualenv` target
-# or the global `preinstall` topic. If the virtualenv has not been
-# preinstalled, it will be created on-demand when needed during the
-# creation of a `FreedomESDKTarget`.
+# the creation of a Virtualenv relative to the BuildDir of Freedom E SDK's
+# Build Environment. By default, these wake rules will ensure the creation
+# of this Virtualenv at run-time with `installFreedomESDKVirtualenv`.
+#
+# If your use-case requires that dependencies are fetched from the internet
+# at workspace creation time, make sure to add `installFreedomESDKVirualenv`
+# to the `preinstall` topic, substituting the FreedomESDKBuildEnv you want
+# to use.
 #
 
-export def freedomESDKBuildDir = "build/freedom-e-sdk"
-export def freedomESDKVenvDir = "{freedomESDKBuildDir}/venv"
+export def freedomESDKVenvDir buildEnv = "{buildEnv.getFreedomESDKBuildEnvBuildDir}/venv"
 
-def makeFreedomESDKVirtualenvJob Unit =
+def makeFreedomESDKVirtualenvJob buildEnv =
   def runDir = here
+  def resources = buildEnv.getFreedomESDKBuildEnvResources
 
   def inputs =
     source "freedom-e-sdk/scripts/virtualenv.mk",
     source "freedom-e-sdk/requirements.txt",
-    mkdir freedomESDKVenvDir,
+    mkdir (freedomESDKVenvDir buildEnv),
     Nil
 
   def cmd =
-    def venvDir = relative runDir freedomESDKVenvDir
+    def venvDir = relative runDir (freedomESDKVenvDir buildEnv)
     "bash", "-c", "%
       export FREEDOM_E_SDK_VENV_PATH=$(pwd)/%{venvDir}
       make -f scripts/virtualenv.mk virtualenv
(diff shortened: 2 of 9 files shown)


build.wake
```diff
@@ -0,0 +1,430 @@
...
+  | filter (!matches `.*\.wake` _.getPathName)
+
+#
+# Plan Resources
+#
+
+global def defaultFreedomESDKResources =
+  "riscv-tools/2019.08.0",
+  "python/python/3.7.1",
+  Nil
```
richardxia, line 68: We don't really have a great solution or general pattern in place for this, but I think we need a better way of allowing this stuff to be overridable. We often name these kinds of variables `defaultFooResources`, but we don't actually provide interfaces for overriding the defaults. One challenge in reusing rules is that we often want to control the concrete tool versions at the top level. I feel that you really ought to be able to say "I require Python 3.5+", and then the top-level can choose to resolve it down to a concrete version, since we may want to minimize the number of different Python versions in use in a single workspace.

In lieu of a longer-term solution, I feel that we need some argument to one of the public functions that lets one actually override these resources.

build.wake
```diff
@@ -0,0 +1,430 @@
...
+# Freedom E SDK depends on a number of Python libraries. Here we enable
+# the installation of a single, workspace-wide virtualenv which can be
+# preinstalled using either the `preinstallFreedomESDKVirtualenv` target
+# or the global `preinstall` topic. If the virtualenv has not been
+# preinstalled, it will be created on-demand when needed during the
+# creation of a `FreedomESDKTarget`.
+#
+
+global def freedomESDKBuildDir = "build/freedom-e-sdk"
+global def freedomESDKVenvDir = "{freedomESDKBuildDir}/venv"
```
richardxia, line 82: Similar to above, I think the top-level projects will want to have more control over where the build files are created. With this approach, there's a bit of a namespacing issue, where we have to come up with unique build directory names. Although naming it after the repository like this more or less prevents conflicts, this may not necessarily present an organization that makes sense to the user of a top-level project.

For example, if I'm running through the Chisel -> Verilog -> SDK -> RTL simulation flow, such as in the [block-pio-sifive](https://github.com/sifive/block-pio-sifive) flow, the user of the system really cares foremost about the primary artifacts produced in each step above (design Verilog files, design-specific SDK, RTL simulator), and not stuff like "here are the .jar files created by each of the Chisel dependencies, each in its own directory directly in `build/`.

I think allowing these to be overridable by the caller will provide greater flexibility for top-level projects to organize their build directories in a way that makes the most sense for that project.

build.wake
```diff
@@ -0,0 +1,430 @@
...
+
+  def cmd =
+    def venvDir = relative runDir freedomESDKVenvDir
+    "bash", "-c", "%
+      export FREEDOM_E_SDK_VENV_PATH=$(pwd)/%{venvDir}
+      make -f scripts/virtualenv.mk virtualenv
+      # Remove e.g. .fuse/31337 from the paths in the virtualenv
+      find ${FREEDOM_E_SDK_VENV_PATH}/bin -type f | xargs sed -i 's%.fuse/[[:digit:]]\+/%%g'
+      # Remove e.g. .build/31337 from the paths in the virtualenv (when Wake doesn't use FUSE)
+      find ${FREEDOM_E_SDK_VENV_PATH}/bin -type f | xargs sed -i 's%.build/[[:digit:]]\+/%%g'
```
richardxia, line 101: Nothing wrong with your code here, as I fully understand the problem here, but this is just a self-remark that we need to fix this so that the FUSE runner resolves paths into the non-FUSE directory, such that the `.fuse` directory is basically invisible to the underlying processes.

build.wake
```diff
@@ -0,0 +1,430 @@
...
+  | getJobStderr
+
+publish preinstall = (preinstallFreedomESDKVirtualenv), Nil
+
+# Wake rules which depend upon the Freedom E SDK virtualenv should
+# add `installFreedomESDKVirtualenv` to their Plan inputs.
+global target installFreedomESDKVirtualenv Unit =
+  makeFreedomESDKVirtualenvJob Unit
+  | runJob
+  | getJobOutputs
```
richardxia, line 124: I have started developed a personal pattern/habit of not having any important APIs (e.g. `global` functions) directly return the `getJobOutputs`. There's two specific reasons I have for this:

1. I'm not a fan of the current behavior of `getJobOutputs` returning a list of a single BadPath on error, since it's easy to forget to check for that, and I feel that we often run into confusing errors due to propagating BadPaths too far without handling them. I think @terpstra is general in agreement with me about removing BadPath eventually and replacing it with `Result Path` (where Path is only a "good" path) in most places in the standard library.
2. This makes it difficult/impossible to make backwards compatible changes to this API in the future. For example, if we later decided that we wanted to return an additional piece of information, we can't do so without changing the type signature of this function. If, however, we had this return something like a `PythonVirtualEnv` `tuple`, then we can add new fields to that tuple without breaking any interfaces, since the type signature would continue to work.

We don't necessarily have to go to 2) for all interfaces, but I think all global/exported interfaces are ones where we should strongly consider adding tuples since it's more painful to break global/exported interfaces.

build.wake
```diff
@@ -0,0 +1,430 @@
...
+
+#
+# Freedom E SDK Installation
+#
+
+global tuple FreedomESDK =
+  global Dir:            String
```
richardxia, line 131: Yes! I love this! I've definitely wanted this to be returned as part of installing the SDK, since the raw list of `Path`s makes it hard to tell which directory is considered to be the "top" directory for the SDK.

nategraff-sifive:
@richardxia Thanks for all your feedback! I don't have a timeline for getting this in, so I'm happy to iterate several times based on feedback. If I didn't want this API to be better than what we currently had, I wouldn't have bothered in the first place!

.github/workflows/wake.yml
```diff
@@ -0,0 +1,34 @@
...
+    env:
+      WIT_WORKSPACE: wit-workspace
+
+    steps:
+      - uses: actions/checkout@v2
+        with:
+          path: wit-packages/freedom-e-sdk
+
+      - name: 'Init Wit Workspace'
+        uses: sifive/wit/actions/wit@v0.13.0
```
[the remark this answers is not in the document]
mmjconolly: Yeah depending on that hash for now sounds reasonable.
Yep on the v0.13.1 release, though (unrelated to this PR) I may spend a bit of time trying get an auto-push-to-pypi working

build.wake
```diff
@@ -0,0 +1,430 @@
...
+  _
+  | filter (!matches `(.*/)*\..*` _.getPathName)
+  | filter (!matches `.*\.wake` _.getPathName)
+
+#
+# Plan Resources
+#
+
+global def defaultFreedomESDKResources =
+  "riscv-tools/2019.08.0",
```
bsousi5, line 66: Is possible to support a list of tools versions?
And also allow caller to specify what tools version to use, similar to choosing VCS vs Verilator.
By default it can use the version define here.
But caller may be able to select a particular version or all of the support tools versions.

nategraff-sifive:
@richardxia @bsousi5 WRT changing resources and build paths:

How about something like:

```
export tuple FreedomESDKBuildEnv =
  export Resources: List String
  export BuildDir: String

def defaultFreedomESDKResources =
  "riscv-tools/2019.08.0",
  "python/python/3.7.1",
  Nil

def defaultFreedomESDKBuildDir = "build/freedom-e-sdk"
def freedomESDKVenvDir buildEnv = "{buildEnv.getFreedomESDKBuildEnvBuildDir}/venv"

export def defaultFreedomESDKBuildEnv =
  FreedomESDKBuildEnv defaultFreedomESDKResources defaultFreedomESDKBuildDir
```

And then the public API accepts FreedomESDKBuildEnv as a parameter. The SDK can store the build environment tuple so callers only need to pass it in once.

```
export def makeFreedomESDK buildEnv destDir =
  ...
```

Callers can override their build environment to something like:

```
def myFreedomESDKBuildEnv =
  defaultFreedomESDKBuildEnv
  | setFreedomESDKBuildEnvBuildDir "some/other/path"
  | setFreedomESDKBuildEnvResources ("riscv-tools/3099.13.32", "python/python/2.7", Nil)
```

.github/workflows/wake.yml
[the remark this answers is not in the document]
nategraff-sifive: Bumped to `wit@9fdc590`, I'll leave this thread unresolved until Wit v0.13.1 is tagged.

build.wake
```diff
@@ -0,0 +1,430 @@
+#
```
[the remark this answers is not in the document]
nategraff-sifive: I'm now using the package system, but I'll need to go back and figure out what the `require` statement does to see where/if I might use that here.

build.wake
nategraff-sifive: I've implemented a fix for this with the FreedomESDKBuildEnv tuple.

build.wake
nategraff-sifive: Added the `FreedomESDKVirtualenv` tuple.

build.wake
nategraff-sifive: <3

build.wake
nategraff-sifive: I've implemented a fix for this with the FreedomESDKBuildEnv tuple.

PR #513, opened 2020-06-29
Remove git fetch no longer need with latest git update

bsousi5:


.github/workflows/release.yml
```diff
@@ -26,12 +26,6 @@ jobs:
           submodules: 'recursive'
```
nategraff-sifive, line 26: Remember to add `fetch-depth: 0` here

PR #521, opened 2020-08-07
Bump FreeRTOS-metal

e-puerto:
Bump FreeRTOS-metal in order to support rv32e architecture


bsousi5:
@e-puerto Can you raise a PR to get this into PCS. Thanks

PR #522, opened 2020-08-10
Fix 'make clean software'

nategraff-sifive:
When building software, make sure that the Freedom Metal codegen is up-to-date.


scripts/standalone.mk
```diff
@@ -254,38 +282,20 @@ PROGRAM_LST ?= $(SRC_DIR)/$(CONFIGURATION)/$(PROGRAM).lst
 .PHONY: all
 all: software
 
+.DEFAULT_GOAL := software
```
keith-packard, line 285: I don't think you'd need this if you hadn't moved the metal.mk generation up in the file. Make scans all the way to the bottom looking for instructions to build included files, so that shouldn't have needed to move?

nategraff-sifive: I did the rearranging, and while I was at it I got rid of scripts/libmetal.mk because there was no purpose to a separate make fragment for that.

PR #527, opened 2020-08-24
Bump FreeRTOS get MTIME ctrl address from platform.h

bsousi5:
This fix is require for configurable base address of CLINT and CLIC


e-puerto (approved):
LGTM

PR #528, opened 2020-08-24
[next] Add Wake API to Freedom E SDK

nategraff-sifive:
Port of #501 to the freedom-metal-next branch.


mmjconolly:
I hope I'm reading this correctly, It looks like there are sections where you find repositories (like example-*), and that the presumption is that wit has checked out the repository in the root of the workspace. This assumption likely wont hold in the not-too-distant future (ie, non-flat workspaces).

Would it be possible to support a repository layout that is either flexible, or alternatively looks like what the git submodules does? For example if we tried to run wake within a federation submodules checkout (which I've been experimenting with lately)
```
$ tree
federation
├── api-chisel3-sifive
├── api-firrtl-sifive
├── ....
├── freedom-e-sdk
│   ├── bsp
│   ├── ....
│   └── software
│       ├── atomics
│       ├── cflush
│       ├── clic-hardware-vector-interrupts
│       ├── ....

$ wake -x 'deliver ...'
```

Happy to talk about this offline if you like

build.wake
```diff
@@ -0,0 +1,416 @@
...
+    sources "elf2hex" `.*`
+    | filterWakeAndDotFiles
+
+  def installedFiles = 
+    copyDir "freedom-e-sdk"                   destDir     sdkSources
+    ++ copyDir "scl-metal"                    sdkDir      sifiveCryptoLibSources
+    ++ installFreedomMetal                    sdkDir
+    ++ copyDir "FreeRTOS-metal"               sdkDir      freeRTOSSources
+    ++ copyDir "Segger_SystemView-metal"      sdkDir      systemViewSources
+    ++ copyDir "elf2hex"                      scriptsDir  elf2hexSources
```
richardxia, line 184: If these are all coming from different wit/wake repos, then I think we should export a constant in each of these packages that holds the path to the root of the repo (which can be computed from `here`). If some of these repos do not contain any Wake code, then I think @mmjconolly has some code he's been working on in some other repos to attempt to locate git repo with a given name even if that repo contains no Wake code.

build.wake
```diff
@@ -0,0 +1,416 @@
...
+      VC707TargetType = "{bspDir}/openocd.cfg", alwaysOutputs
+      VCU118TargetType = "{bspDir}/openocd.cfg", alwaysOutputs
+      HiFiveTargetType = "{bspDir}/openocd.cfg", alwaysOutputs
+      _ = alwaysOutputs
+
+  def installedFiles =
+    makePlan cmdline inputs
+    | setPlanResources buildEnv.getFreedomESDKBuildEnvResources
+    | setPlanFnOutputs fnOutputs
+    | setPlanLocalOnly True
```
richardxia, line 280: Do you know why this is a LocalOnly job? We've improved the sandboxing story in Wake 0.19, including finally fixing that annoying `.fuse/` directory leaking into the absolute paths. If possible, I'd like to eliminate almost all of the LocalOnly jobs, since they break out of the safety mechanisms of Wake (and they tend to be the cause of the instability or non-reproducibility of our builds today).

build.wake
```diff
@@ -0,0 +1,416 @@
...
+    def coreDTS =
+      installedFiles
+      | filter (\f matches `.*core\.dts` f.getPathName)
+      | head
+      | getOrElse (makeBadPath (makeError "Unable to get generated core.dts"))
+    makeFreedomMetalCodeGenerationOptions designDTS (coreDTS, Nil) bspDir
+    | runFreedomMetalCodeGeneration sdkDir
+    | getJobOutputs
+
+  FreedomESDKTarget sdk targetName targetType (installedFiles ++ generatedMetalSources)
```
richardxia, line 308: Similar to my earlier comment, I think we should have this return a `Result FreedomESDKTarget Error` rather than letting `BadPath`s propagate. For this function in particular, I think the new `require` keyword will make this a lot easier to write code for, since before the `require` keyword, we needed to heavily nested pattern matches or pipelines.

I'm not sure if we have any good examples of `require` in the open source code, but I could point you to a few examples in internal code if you'd like.

build.wake
```diff
@@ -0,0 +1,416 @@
...
+  def hexResult = getFile `.*\.hex`
+
+  match elfResult hexResult
+    (Some elf) (Some hex) = Pair elf hex
+    _ _ =
+      def badPath =
+        "{targetName}: Unexpected software ouputs: {format outputs}"
+        .makeError
+        .makeBadPath
+      Pair badPath badPath
```
richardxia, line 378: Another place where I would recommend `Result` over `BadPath`.

build.wake
```diff
@@ -0,0 +1,416 @@
...
+    ++ targetInstall
+    ++ programInstall
+    ++ (installFreedomESDKVirtualenv buildEnv).getFreedomESDKVirtualenvInstalledFiles
+
+  def fnOutputs _ = files "{sdkDir}/software/{programName}/release" `[^/]*.(elf|hex)`
+
+  makePlan cmdline inputs
+  | setPlanResources buildEnv.getFreedomESDKBuildEnvResources
+  | setPlanFnOutputs fnOutputs
+  | setPlanLocalOnly True
```
richardxia, line 414: Once again, do you know why this requires LocalOnly?

programs.wake
```diff
@@ -0,0 +1,119 @@
...
+
+export def getFreedomESDKProgramSourceDir program =
+  def name = program.getFreedomESDKProgramName
+  def kind = program.getFreedomESDKProgramKind
+  "{kind}-{name}"
+
+export def makeFreedomESDKProgram name kind =
+  def program = FreedomESDKProgram name kind Nil
+  def sourceDir = program.getFreedomESDKProgramSourceDir
+  def programSources = sources sourceDir `.*`
```
richardxia, line 18: This makes the assumption that the source directories are at the root of the wit/wake workspace, which may be problematic in the future, like I alluded to above.

test-wake-api.wake
```diff
@@ -0,0 +1,54 @@
+package freedom_e_sdk_test
```
richardxia, line 1: 👍 for writing a Wake test!

test-wake-api.wake
```diff
@@ -0,0 +1,54 @@
...
+    # map to List Result and reduce to a Result
+    map getPathResult programELFs
+    | findFail
+
+  # Pretty-print results
+  match results
+    Pass paths =
+      def _ = map (\p println p.getPathName) paths
+      Pass "All ELFs built successfully for {targetName}"
+    Fail error = Fail error
```
richardxia, line 54:
```suggestion
  require Pass paths = results
  def _ = map (\p println p.getPathName) paths
  Pass "All ELFs built successfully for {targetName}"
```

Here's a simple example of how you can use the new `require` keyword to replace certain kinds of `match` statements that you had to do previously. Note that when `results` is _not_ a `Pass`, the function will fail early with that failure type, which works out because the error portion of the return type of the function matches the this intermediate result's failure type.

richardxia (commented):
Sorry that it took so long for me to get around to reviewing this kind of stuff. Thanks for being an early adopter of a bunch of Wake stuff! I'm happy to discuss any parts of my feedback in more detail, especially the Wit workspace layout-related things.

build.wake
```diff
@@ -0,0 +1,416 @@
...
+
+  def cmd =
+    def venvDir = relative runDir (freedomESDKVenvDir buildEnv)
+    # Explicitly create the venv with the 'python3.7' binary so that we use
+    # the version of python with pip and venv preinstalled in all build environments.
+    "bash", "-c", "%
+      export FREEDOM_E_SDK_VENV_PATH=$(pwd)/%{venvDir}
+      python3.7 -m venv ${FREEDOM_E_SDK_VENV_PATH}
+      . ${FREEDOM_E_SDK_VENV_PATH}/bin/activate && python3 -m pip install pip==20.0.1 --disable-pip-version-check
+      . ${FREEDOM_E_SDK_VENV_PATH}/bin/activate && python3 -m pip install -r requirements.txt
```
mmjconolly, line 120: This superficially seems like its doing something similar to https://github.com/sifive/api-languages-sifive/blob/master/python.wake , is there enough overlap to make use of python.wake?

build.wake
```diff
@@ -0,0 +1,416 @@
...
+  export InstalledFiles: List Path
+
+export def freedomESDKVenvDir buildEnv = "{buildEnv.getFreedomESDKBuildEnvBuildDir}/venv"
+
+def makeFreedomESDKVirtualenvJob buildEnv =
+  def runDir = here
+
+  def inputs =
+    source "freedom-e-sdk/scripts/virtualenv.mk",
+    source "freedom-e-sdk/requirements.txt",
```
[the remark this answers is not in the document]
nategraff-sifive: Thanks, I'll make that change.

.github/workflows/wake.yml
```diff
@@ -0,0 +1,26 @@
...
+    steps:
+      - name: 'Init Wit Workspace'
+        uses: sifive/wit/actions/wit@v0.13.1
+        with:
+          command: init
+          arguments: |
+            ${{ env.WIT_WORKSPACE }}
+            -a git@github.com:sifive/freedom-e-sdk.git::${{ github.sha }}
+            -a git@github.com:sifive/environment-blockci-sifive.git::0.7.1
+          force_github_https: true
```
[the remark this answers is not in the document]
nategraff-sifive: I used that, but had to switch back to this so I could put a `-v` in to debug it. I can switch back to the init action now.

I'll also make the update to the 0.8.8 blockci env.

build.wake
```diff
@@ -0,0 +1,416 @@
...
+# Freedom E SDK Installation
+#
+
+export tuple FreedomESDK =
+  export Dir:            String
+  export InstalledFiles: List Path
+  export BuildEnv: FreedomESDKBuildEnv
+
+# makeFreedomESDK assembles the components of Freedom E SDK into the directory
+# hierarchy expected by the SDK from their flat layout in the wit workspace.
```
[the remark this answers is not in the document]
nategraff-sifive: This particular rule allows you to build another freedom e sdk somewhere else, like in an IP delivery.

build.wake
nategraff-sifive: Not sure, I think it's carried over from the original wake for this task. I suppose I'll get rid of it and find out if it's necessary.

build.wake
nategraff-sifive: Yeah, at this point I think it's fine to just depend on the api-languages-sifive version.

.github/workflows/wake.yml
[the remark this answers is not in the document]
richardxia: @terpstra, we might want to add an option to the `wit/init` GitHub Action that allows one to set the verbosity level.

build.wake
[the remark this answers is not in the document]
richardxia: Right, I was probably unclear in my wording, but I actually meant to comment on the location of the source files, not the destination files. We're making a push make the Wake rules agnostic of the location of the Git repositories, so it will not be safe to assume that all Git repos are at the workspace root.

PR #542, opened 2020-09-15
Remove picolibc support

nategraff-sifive:


paul-walmsley-sifive (approved):
Looks reasonable, but I'm not an expert in this codebase.  The general ideal is aligned with SiFive planning

PR #546, opened 2020-09-24
Update README with correct default TARGET

knightsifive:
On my workstation, this little gemstone
```
TARGET ?= $(shell find $(TARGET_ROOT)/bsp/* -type d | head -n 1 | rev | cut -d '/' -f 1 | rev)
```
currently selects `freedom-e310-arty`, contradicting README.md.

This behavior is apparently determined by the current population of the `bsp` subdirectory, so future revisions of FESDK could inadvertently re-break this. (I suspect this behavior is also shell dependent.)

### DIFF
--- a/README.md
+++ b/README.md
@@ -296,12 +296,12 @@ export FREEDOM_E_SDK_VENV_PATH=/path/to/venv
 To compile a bare-metal RISC-V program:
 
 ```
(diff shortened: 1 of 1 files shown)


nategraff-sifive (approved):
LGTM

PR #548, opened 2020-09-25
SDS1 Fix memory overrun in l2pf, l2pm examples.

NandkumarJoshi:
Updated examples, to limit memory under test.

### DIFF
--- a/wit-manifest.json
+++ b/wit-manifest.json
@@ -5,7 +5,7 @@
         "source": "git@github.com:sifive/freedom-devicetree-tools.git"
     },
     {
-        "commit": "d47cd29a097416b8747938210a6e76d84c29b982",
+        "commit": "7150fb3619a46fbb1f8d8a8df4f90e4e0ee0fccd",
         "name": "freedom-metal",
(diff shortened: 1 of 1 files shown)


NandkumarJoshi:
tested this fe-sdk commit to run OK against pcs-kiwi release branch.

wit-manifest.json
```diff
@@ -260,12 +260,12 @@
         "source": "git@github.com:sifive/example-hca-metal.git"
     },
     {
-        "commit": "950b3ac73c3e5a2cf7e7463aa05b144e8616569c",
+        "commit": "ea10676cde77d8417b0b805b32631708f2f72a91",
```
bsousi5, line 263: For both examples, add a comment the purpose of +16 for the address, and use metal api for cflush for consistency. you can put the asm in a comment as an alternate usage.

NandkumarJoshi: updated examples with comment, metal API for cache flush.

bsousi5 (approved):
LGTM

PR #549, opened 2020-09-25
Bump FreeRTOS to fix rv32d

nategraff-sifive:


paul-walmsley-sifive (approved):
I don't know the technical background behind these fixes, so I can't provide any technical review here.  But I trust Nate and EmmanuelP on this one

PR #550, opened 2020-09-28
Update release workflow for fetch depth

nategraff-sifive:


paul-walmsley-sifive:
Nate writes that this is a very low impact change, that we've already done this on other branches, and that without it the Github release automation doesn't work.  Nate, no need to ask for review on these kinds of changes; just go ahead and merge them.

PR #553, opened 2020-10-03
Fix l2pf base address defines.

NandkumarJoshi:


bsousi5:
Is there an update to the BSP files. or may be not because QEMU, as they would not have L2PF

bsousi5 (approved):
LGTM

PR #559, opened 2020-10-14
Removed submodule Segger_SystemView-metal

e-puerto:
Segger_SystemView is no longer used in FreeRTOS, so we remove it from Freedom-e-sdk


nategraff-sifive (approved):
LGTM

PR #560, opened 2020-10-19
[next] Fixes for running in CoreIP RTL Simulation

nategraff-sifive:


keith-packard (approved):
I'd probably switch METAL_WAIT_CYCLE to just WAIT_CYCLE and remove the #ifndef as we really want to ignore anything set from outside the two interrupt examples. Also, should you use `#if __riscv_flen` to determine if there's an FPU rather than 'riscv_fdiv'?

keith-packard (approved):
I'd probably switch METAL_WAIT_CYCLE to just WAIT_CYCLE and remove the #ifndef as we really want to ignore anything set from outside the two interrupt examples. Also, should you use `#if __riscv_flen` to determine if there's an FPU rather than 'riscv_fdiv'?

knightsifive:
I second @keith-packard's suggestion, although I don't see what file he's referencing. My understanding, based on the (very rough) draft [C API](https://github.com/riscv/riscv-c-api-doc/blob/master/riscv-c-api.md#preprocessor-definitions), is that `__riscv_flen` is currently the most portable way of checking for the presence/absence of the F, D, or Q extensions.

PR #563, opened 2020-10-28
add 'list-freedomstudio-helpers' target

sifivekevin:
Adding manually managed meta information that Freedom Studio can use to ensure the best possible out-of-box experience when create new example software projects.  Discuss and review!

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -206,7 +206,6 @@ freedomstudio-api-version:
 #
 # compiler = gcc 
 # compiler_version = X[.Y[.Z[]]]
-# add_fcommon = (true|false)
 #
 # Only compiler=gcc is supported right now.
 #
--- a/Makefile
+++ b/Makefile
@@ -192,10 +192,33 @@ clean:
 # A given version of Freedom Studio may not support 
 # 'future' versions of freedom-e-sdk.  Backward support
 # should be supported.
+#
+# 1.0 - Original API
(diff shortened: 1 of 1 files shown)


nategraff-sifive (approved):
LGTM

Makefile
```diff
@@ -192,10 +192,33 @@ clean:
...
+#
+# compiler_version: can be as general or specific as necessary
+# if omitted, then any or only compiler is used. If specified
+# compiler is not present then FS will display a message to
+# the user.
+#
+#############################################################
+.PHONY: list-freedomstudio-helpers
+list-freedomstudio-helpers:
+	@echo compiler=gcc, compiler_version=10
```
bsousi5, line 221: In the comment, there is add_fcommon variable, that is not needed? If needed, what select true or false? is it compiler_version?

sifivekevin: Just removed this.  Was thinking about adding that, but have decided not to. There are (or will be) better ways of handling this.

bsousi5 (approved):
Thinking more about it, these variables (compiler and compiler_version) may need to be a list. Unless for version it is imply that its 8.3 and 10.1. 
Will leave to software team to decide how to tackle this in the future, but IMO, it is better to be explicit something like = {gcc, llvm} and = {[8.3, 10.1], []}.

PR #572, opened 2020-11-25
[next] Add support for L2 LIM/Prefetcher/Performance Monitor

nategraff-sifive:
- Bumps metal for L2 features
- Adds example-lim, example-l2pf, example-l2pm
- Regenerates linker scripts for L2 LIM support


keith-packard:
I'll let you get these passing the tests; I don't see much to review here though, just a few new examples?

PR #578, opened 2020-12-29
Add example for privilege switch, include S-mode and U-mode

zongbox:
Here is the repo of this example program: 
https://github.com/sifive/example-privilege-level

As README.md said, this program demonstrates how to switch mode to supervisor mode from machine mode, then switch mode to user mode from supervisor mode.

This program shows what should be done before switching mode, and uses ecall to test mode transfer. After switching mode, we call the ecall and check if we get the correct corresponding exceptions.
- Call ecall in S-mode, get the exception code 9 (Environment call from S-mode)
- Call ecall in U-mode, get the exception code 8 (Environment call from U-mode)

Test on QEMU:
https://github.com/sifive/qemu-private/commits/dev/yihaoc/bare-s 
and  
https://github.com/sifive/qemu-private/commits/dev/zongl/bare-s-finisher  (with test finisher)

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -157,3 +157,6 @@
 [submodule "software/example-lim"]
 	path = software/example-lim
 	url = https://github.com/sifive/example-lim.git
+[submodule "software/example-privilege-level"]
+	path = software/example-privilege-level
+	url = https://github.com/sifive/example-privilege-level.git
--- a/wit-manifest.json
+++ b/wit-manifest.json
@@ -263,5 +263,10 @@
         "commit": "cd04f06675b3ba464ffabd72e5800b9ba1313104",
         "name": "example-lim",
         "source": "git@github.com:sifive/example-lim.git"
+    },
+    {
+        "commit": "e296353a93f816d1f9ddbc8898b3014c8686c4ce",
+        "name": "example-privilege-level",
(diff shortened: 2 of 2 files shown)


zongbox:
@bsousi5  @paul-walmsley-sifive We're going to put the example programs in freedom-e-sdk directly instead of submodule, so I close this PR first, and then sent an another PR to do that later. Thanks.

knightsifive:
@zongbox thanks for the example. Just so you know, Freedom-E-SDK is a public repository, but your PR includes links to a "qemu-private" repo, which appears to be a private SiFive repo.

bsousi5:
Thanks @zongbox. Can you also hold off on creating the PR by a day, will update you on the plan for freedom-e-sdk on by EOB

zongbox:
@knightsifive This program works on public qemu as well, but public qemu doesn't support sifive finisher, so it doesn't exit normally.

I have another example for bare S mode testing, the satp and sfence behavior are different in bare S mode, these changes are only implemented in qemu-private now. I was wondering that the policy about example programs of FESDK, and is it OK to put the bare S mode example in FESDK? OR we should wait qemu upstream first?

Thanks.

zongbox:
@bsousi5 Yes, thanks for the notification. I would postpone PRs until that.

PR #581, opened 2021-01-15
Update sifive-welcome.c

sifivekevin:
removing extra 'and' from code comment

Not sure who the official maintainer is right now, so @richardxia get to approve/merge or delegate this.


richardxia:
@paul-walmsley-sifive is the official maintainer currently.

PR #585, opened 2021-02-19
add semihost library support

sifivekevin:
GCC-10.2.0-2020.12.8 include a functional semihost library.  This change allows freedom-e-sdk standalone projects to build example using this library by providing `RISCV_LIBC=semihost` 

This has been tested in Freedom Studio.  Freedom Studio only offers the `semihost` option if the chosen toolchain has a semihost library present in the installation.  

Snag_4fbe006

### DIFF
--- a/scripts/standalone.mk
+++ b/scripts/standalone.mk
@@ -60,6 +60,13 @@ METAL_WITH_EXTRA=--with-builtin-libgloss
 SPEC=nano
 endif
 
+ifeq ($(RISCV_LIBC),semihost)
+# semihost must come before metal-gloss
+LIBMETAL_EXTRA=-lsemihost -lmetal-gloss 
+METAL_WITH_EXTRA=--with-builtin-libgloss
+SPEC=semihost
+endif
+
 ifeq ($(SPEC),)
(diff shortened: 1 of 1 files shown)


SiFiveGregS (approved):
Verified that the makefile snippet generates a working semihosting configuration (using Arty 100T and a recent IP package and with semihost selected as RISCV_LIBC)

keith-packard:
Does this use the semihost support recently merged to newlib? I wouldn't think this has anything to do with gcc, but I haven't looked to see if there is semhosting support there as well?

sifivekevin:
When I said "GCC", I mean everything that is included in the toolchain package.  In this case of the latest 2020.12.8 package I simply look for this file to indicate to me that semihost is supported in the toolchain package.:

`<toolchain-root>\riscv64-unknown-elf\lib\libsemihost.a`

e-puerto:
> `<toolchain-root>\riscv64-unknown-elf\lib\libsemihost.a`

Could we know where are the source of this library? I have merged 2 days ago the modification into freedom-metal (libgloss) in order to support the semihosting (PR that I have created 1 year ago)

PR #588, opened 2021-03-23
software: Add a example for illustrating privilege transfer

zongbox:
As README.md said, this program demonstrates how to switch mode to supervisor mode from machine mode, then switch mode to user mode from supervisor mode.

This program shows what should be done before switching mode, and uses ecall to verify mode transfer. After switching mode, we call the ecall and check if we get the correct corresponding exceptions.

- Call ecall in S-mode, get the exception code 9 (Environment call from S-mode)
- Call ecall in U-mode, get the exception code 8 (Environment call from U-mode)

Test on QEMU:
https://github.com/sifive/qemu-private/commits/dev/zongl/bare-s-finisher (with test finisher)

### DIFF
--- a/software/example-privilege-level/LICENSE
+++ b/software/example-privilege-level/LICENSE
@@ -1,0 +1,3 @@
+This source repository is release under Apache2 and MIT licenses.
+
+See LICENSE.Apache2 and LICENSE.MIT for details.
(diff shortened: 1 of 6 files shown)


zongbox:
Hi Emmanuel, I was wondering if you can help me to review this patch and merge it if it is good enough to you. Thanks a lot.

e-puerto (approved):
LGTM

PR #598, opened 2021-05-12
doc: fix an error if the doc is not generated from github

sbourdelin:
The commit 57c45c04 expect Sphinx to find the environment variable
RELEASE_TAG.
It seems it's only the case when the documentation is generated from
github using the workflow defined in the release.yml file.

This commit allows to generate the doc from outside github, by
retrieving the ref or tag of the current branch if we are on a git
repository.
Otherwise the version is set to "n/a".


e-puerto (approved):
LGTM

PR #601, opened 2021-06-02
Add SiFive Stride Prefetcher driver on qemu-sifive-e31-customer platform

frankchang0125:
Add SiFive Stride Prefetcher (SPF) driver.

To build example (any example is available, as long as SPF DTS node is included in the selected BSP. Here is `qemu-sifive-e31-customer` BSP):
`make PROGRAM=hello TARGET=qemu-sifive-e31-customer CONFIGURATION=release software`

QEMU with SPF support: https://github.com/sifive/qemu/tree/sifive-e-customer-v6.0.0

To run example with QEMU:
`qemu-system-riscv32 -nographic -M sifive_e_customer -bios none -kernel <your path to example>/hello.elf`

`sifive_l2pf1_init()` will be called during boot up if SPF node is included in DTS (`sifive,l2pf1`), which will initialize SPF with predefined values.

Multi-core SPF is also available in `qemu-sifive-e31mc-customer` BSP.


e-puerto (approved):
LGTM

PR #617, opened 2021-08-03
Add build dirs to .gitignore

vbendeb:
These are directories generated when building various example images.

TEST='git status' does not show generated directories any more


ManuelonGithub:
matters little since pull requests don't seem to be reviewed, but consider adding `software/*/release/` as well

PR #620, opened 2021-09-03
Add BSP for SparkFun RED-V board

danak6jq:
Some time ago, I took a stab at adding a BSP for the SparkFun RED-V, similar to the HiFive B. I've used the BSP for a few projects but openly admit I was stumbling around in dim light when I did this. I suggested to SparkFun they might want it, but don't recall hearing back.

Opening a PR to (a) invite expert review and (b) perhaps contribute it upstream.


chaoyangnz:
sad, no response?

ManuelonGithub:
Very sad indeed. I am going through the changes you've made and I experienced errors when building with a new clone of the freedom-e-sdk. I've concluded that the new BSP name (i.e. the directory name) must have "hifive1-revb" in its name. Otherwise you will get the following undesired behaviour when building inside of freedom-e-sdk:

- [generate_overlay.py](https://github.com/sifive/devicetree-overlay-generator/blob/89be54a0e0092d965e87661035fbc02a58292c48/generate_overlay.py) will throw an error because it must see "hifive" somewhere in the target name.
- [generate_settings.py](https://github.com/sifive/esdk-settings-generator/blob/823f838b468edb0ef720427a73f9b30eb864ff6a/generate_settings.py) will overwrite bsp/{TARGET}/settings.mk to have "TARGET_TAGS=board openocd" because it must see "hifive1-revb" somewhere in the target name to write "TARGET_TAGS=board jlink" instead.

Although I am not a fan of how large it forces the name to be (my current fork has it as sparkfun-redv-hifive1-revb), the alternative would be to modify the two submodules and make a pull request on each... I would stick with the longer name route.

PR #634, opened 2022-03-06
Added core cycle count, core cycles/iter, and CoreMarks/MHz

BrianGraysonSiV:
This patch does not change any of the existing time-reporting, but adds in the core cycle counts, and a computation of CoreMarks/MHz. This makes it easier for customers who do not alter the environment to set their RTC tick frequency #define properly.

I have tested this on E76 for Xilinx, but of course additional testing would be great.

### DIFF
--- a/software/coremark/core_main.c
+++ b/software/coremark/core_main.c
@@ -32,6 +32,24 @@ Original Author: Shay Gal-on
 	Returns:
 	NULL.
 */
+
+#include <metal/csr.h>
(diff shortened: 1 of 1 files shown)


avernhetS5:
@BrianGraysonSiV 
Thanks for sharing.
2 things about this PR:
- it has been done against FESDK public which is not really maintained anymore
- it modifies a file that is supposed to stay untouched AFAIU, since it is a common reference to share results

